Issue #8, opened 2019-11-28
Maps navigation entry


skjnldsv:
Add link to maps if installed or the maps in the store if not installed

jakobroehrl:
There is already a link called Locations, is this the right one?
How can I find out, If an app is installed or not?

skjnldsv:
Yes this is the one :)
I think you will have to add the check in phph and send the ingo to the front by using the initialstate api

There is already one example of it in the app, you could create a second data set for the map?

jakobroehrl:
> ... send the ingo to the front by using the initialstate api

What do you mean by that?

> There is already one example of it in the app, you could create a second data set for the map?

Could't find it. Would you give me a hint? 😬

skjnldsv:
https://github.com/nextcloud/photos/blob/893b8974a091af019681cd7669133287edac5655/lib/Controller/PageController.php#L70
https://github.com/nextcloud/photos/blob/893b8974a091af019681cd7669133287edac5655/src/services/AllowedMimes.js#L25

PR #13, opened 2019-11-29
Maps navigation entry

jakobroehrl:
https://github.com/nextcloud/photos/issues/8
Is this going into the right direction?

Right now I get an error: 
Undefined class constant 'MAPS'


lib/Controller/PageController.php
```diff
@@ -68,12 +68,12 @@ public function index(): TemplateResponse {
 		$this->eventDispatcher->dispatch(LoadViewer::class, new LoadViewer());
 
 		$this->initialStateService->provideInitialState($this->appName, 'mimes', Application::MIMES);
+		$this->initialStateService->provideInitialState($this->appName, 'maps', Application::MAPS);
```
skjnldsv: `Application::MAPS` doesn't exists,
you need to fetch the proper data for it.
You can check if an app is enabled by using this
`$this->config->getAppValue('maps', 'enabled', 'no') === 'yes')`
Make sure you inject the config dependency in this file's constructor as well

src/Photos.vue
```diff
@@ -22,6 +22,7 @@
 
 <template>
 	<Content app-name="photos">
+		{{MapsInstalled}}
```
skjnldsv: you cannot use variables that are not in computed/data in the template.
Also please use variables name like `isMapsInstalled` as it's more straightforward :)

lib/Controller/PageController.php
jakobroehrl: Sorry, no idea what to do here. 
I don't know where to add "$this->config->getAppValue('maps', 'enabled', 'no') === 'yes')" 
and what "Make sure you inject the config dependency in this file's constructor as well" could mean...

skjnldsv: Are you familiar with our nextcloud app ecosystem? :)
We updated our tutorial recently here: https://docs.nextcloud.com/server/stable/developer_manual/app/tutorial.html

the provideInitialState requires 3 argments, the name of the app, the unique key of the data set and the data you want to pass.
Here we want to pass if the maps is enabled or not (boolean).
`Application::MAPS` doesn't exists. Look at the Application.php file.
https://github.com/nextcloud/photos/blob/master/lib/AppInfo/Application.php

skjnldsv: So you need to pass the data we want, meaning `$this->config->getAppValue('maps', 'enabled', 'no') === 'yes')` that will return true/false if maps is enabled.

BUT $this->config is not injected here
https://github.com/nextcloud/photos/blob/c73e919829fd43d0f133609a8457790d447baa13/lib/Controller/PageController.php#L47-L55
You need to import [the IConfig interface](https://docs.nextcloud.com/server/stable/developer_manual/api/OCP/IConfig.html) the same way we inject the IRequest or IEventDispatcher ;)

Then you will be able to pass `$this->config->getAppValue('maps', 'enabled', 'no') === 'yes')` as the maps initial state

skjnldsv: ANd the nextcloud php ecosystem will do the rest and give you access to `$config`, which you can then assign to `$this->config` and use it :)

src/Photos.vue
skjnldsv: @jakobroehrl ^

jakobroehrl:
Thanks for you detailed description. Understand a little bit more :)
So, now I got the state, but have problems to display it:
Uncaught Error: Could not find initial state undefined of maps
    at loadState (index.js:14)
    at Module../src/services/MapsInstalled.js (MapsInstalled.js:24)

lib/Controller/PageController.php
skjnldsv: Well done :)

src/Photos.vue
jakobroehrl: will change it, when it's working. 
There is still one error:
Uncaught Error: Could not find initial state undefined of maps
at loadState (index.js:14)
at Module../src/services/MapsInstalled.js (MapsInstalled.js:24)

lib/Controller/PageController.php
```diff
@@ -47,12 +48,14 @@ class PageController extends Controller {
 	public function __construct($appName,
 								IRequest $request,
 								IEventDispatcher $eventDispatcher,
+								IConfig $config,
 								IInitialStateService $initialStateService) {
-		parent::__construct($appName, $request);
+		parent::__construct($appName, $request, $config);
```
skjnldsv: This is not needed here. :)
The construct of the parent (Controller, see line 38) doesn't needs $config

skjnldsv: https://github.com/nextcloud/server/blob/master/lib/public/AppFramework/Controller.php#L74

src/Photos.vue
skjnldsv: loadstate() require the app name as first argument https://github.com/nextcloud/photos/blob/master/src/services/AllowedMimes.js#L25

jakobroehrl: > 
> 
> loadstate() require the app name as first argument https://github.com/nextcloud/photos/blob/master/src/services/AllowedMimes.js#L25

Found this 2mins before by myself, but it doesn't work neither

skjnldsv: It works for me :)

jakobroehrl: OK, now it's working. :)

jakobroehrl:
Last question: How should I change the link here to navigate correctly?

- https://github.com/nextcloud/photos/blob/74120cc0e33eff540104a7912d05d896ab430bcd/src/Photos.vue#L35
- https://github.com/nextcloud/photos/blob/74120cc0e33eff540104a7912d05d896ab430bcd/src/Photos.vue#L74-L80

src/Photos.vue
```diff
@@ -69,10 +68,18 @@ export default {
...
-			MapsInstalled
+			videoplaceholder
 		}
 	},
+	computed: {
+		mapsLink() {
+			if (isMapsInstalled) {
+				return 'apps/maps'
+			} else {
+				return 'https://apps.nextcloud.com/apps/maps'
```
skjnldsv: I'm wondering if we should rather redirect to the nextcloud apps settings ?
https://nc/settings/apps/integration/maps

src/Photos.vue
```diff
@@ -67,9 +68,18 @@ export default {
...
 			svgplaceholder,
 			imgplaceholder,
-			videoplaceholder,
+			videoplaceholder
 		}
 	},
+	computed: {
+		mapsLink() {
+			if (isMapsInstalled) {
+				return 'apps/maps'
```
skjnldsv: You should return thesame as the original object: `{name: 'maps'}`

jakobroehrl: 1. How can I get the individual server path? http(s)://host/path/index.php
2. How do I have to change the router to navigate to the new route? It appends the route path only on the existing route in the browser

Thanks!

src/Photos.vue
jakobroehrl: changed it

src/Photos.vue
skjnldsv: you need to use the generateUrl method for both.
Ignore my first comment about `{name: 'maps'}` I got confused about the old code :)
What you did was good, but you need to use the generateUrl.

skjnldsv: https://github.com/nextcloud/photos/blob/b2c8850d058551e71fa672e7c0002ef4f02ca65c/src/router/index.js#L23
https://github.com/nextcloud/photos/blob/b2c8850d058551e71fa672e7c0002ef4f02ca65c/src/router/index.js#L37

jakobroehrl:
What do you think about this approach?
https://github.com/nextcloud/photos/blob/903091675cb3e6ba7c4779650fcf82413c8167f5/src/router/index.js#L86-L95

Sill have the problem, there is no redirect, it only appends to the URL:
grafik

src/router/index.js
```diff
@@ -84,8 +84,13 @@ export default new Router({
 			}),
 		},
 		{
-			path: '/maps',
-			name: 'maps',
+			path: generateUrl('apps/maps', ''),
```
skjnldsv: you can use the generateurl directly in the Photos.vue file, there is no need to create new routes :)

src/Photos.vue
```diff
@@ -70,6 +71,15 @@ export default {
 			videoplaceholder,
 		}
 	},
+	computed: {
+		mapsLink() {
+			if (isMapsInstalled) {
+				return {name: 'MapsInstalled'}
```
skjnldsv:
```suggestion
				return generateUrl('apps/maps')
```

src/Photos.vue
```diff
@@ -70,6 +71,15 @@ export default {
...
+			} else {
+				return {name: 'MapsNotInstalled'}
```
skjnldsv:
```suggestion
				return generateUrl('/settings/apps/integration/maps')
```

src/router/index.js
skjnldsv: See other comments :)

src/Photos.vue
skjnldsv: (don't forget to import the generateUrl method on this file too :) )

src/router/index.js
```diff
@@ -84,13 +84,9 @@ export default new Router({
 			}),
 		},
 		{
+			path: '/maps',
```
skjnldsv: you should be able to also drop this entry :)

jakobroehrl:
Hey @skjnldsv I changed the routes. 
But still have this problem:
https://user-images.githubusercontent.com/24530680/70143349-a27c0480-169b-11ea-8278-739fa32d0e74.png

skjnldsv:
> But still have this problem:
> [user-images.githubusercontent.com/24530680/70143349-a27c0480-169b-11ea-8278-739fa32d0e74.png](https://user-images.githubusercontent.com/24530680/70143349-a27c0480-169b-11ea-8278-739fa32d0e74.png)

Ah right! :/
We indeed do not support external url with vue router
Let me manage this  :)

skjnldsv:
Pushed!
@jakobroehrl please review :)

PR #25, opened 2019-12-05
Shift navigation on mobile view

skjnldsv:
Capture d’écran_2019-12-05_16-41-38

Looks weird because of the spacing the navigation have, any idea @jancborchardt ?


src/components/Navigation.vue.orig
```diff
@@ -0,0 +1,209 @@
...
+	&__title {
+		margin: 0;
+	}
+<<<<<<< HEAD
+	&__share {
+		margin-left: 10px;
+=======
+	&--mobile {
+		margin-left: 50px;
+>>>>>>> Shift navigation on mobile view
```
jancborchardt: Some merge conflict here?

jancborchardt:
> Looks weird because of the spacing the navigation have, any idea @jancborchardt ?

Yup, just increase the padding of the menu icon to vertically align with the back arrow. :) Probably also needs a bit of a shift to the right.

I could do it, but the current pull request seems to be conflicting?

src/components/Navigation.vue.orig
skjnldsv: We can remove this file, wrongly commited

skjnldsv:
> I could do it, but the current pull request seems to be conflicting?

now should be fine :+1:

jancborchardt:
@skjnldsv btw, we did say that in subfolders, there should only be the back arrow, _not_ the menu button. So basically, in the root folder, simply change the icon of the back arrow button, and make it open the navigation instead of going back. :)

We can force hide the navigation-toggle which gets inserted globally.

skjnldsv:
Ah right, we can start doing this here! :rocket: 
We need to think about a more standard way of doing this in the vue components than manually hiding it in apps :/
Can you open an issue there? We should discuss the technical way :+1:

jancborchardt:
Before a standard is formed, we first try how it works in an app, like here. :) Would we really manage to standardize this for 18?

Also no idea about the technical way. I guess it should be an action which either goes up one level, or if in root it opens the navigation?

skjnldsv:
> Would we really manage to standardize this for 18?

No, but we can do both? :p
Do it here and discuss standards on nc-vue?

Mikescops:
Will be fixed in #468

skjnldsv:
> Will be fixed in #468

AH, now I get it.
Please don't mix pr, especially with the complexity of 468 :pray:

Mikescops:
@skjnldsv I fixed unintentionally, stars aligned when i moved the Navigation component to put the new grid ahah

Mikescops:
As said, fixed by #468 
image

Issue #6, opened 2019-11-18
Albums design review


jancborchardt:
Just looked through the app (on the albums branch) and noted down some of the possible issues and enhancements:

- [x] Heading in "Your albums" view says "Photos", should be "Your albums" (and for other sections that respective section :)
- [x] The font-size of the album cover texts is a bit big, it’s the equivalent of 19px. That’s almost the 20px of the h2. Better would be something like 16px
- [x] The breakpoint at 1024 when the navigation shows again makes the images look really small. Goes from 5 pictures without nav → 6 pictures _with_ nav. Instead should be 5 pics without → 4 pics with nav?
- [ ] When you zoom the interface it’s a bit strange, as when you zoom in actually the pictures get smaller? This seems to be cause the navigation gets bigger but the pictures stay the same amount and thus get smaller?
- [x] On mobile, the heading overlaps with the menu icon. Heading should be moved to the right accordingly → only on mobile. Also the menu icon could be moved down a bit to vertically align with the heading
- [x] On mobile in subfolders, the back button and menu icon overlap. When in a subfolder, only the back button should show and not the menu icon
- [x] Back button in empty folders seems missing still? Also should show the title as heading for reference
- [x] The emptycontent text says "This folder does not contain pictures", better "No photos in here" (like "No files in here" in Files) much shorter
- [ ] (The Undraw emptycontent graphic looks a bit out of place. Very round etc, where the Photos app is blocky and sleek. Not sure if this is an issue)
- [x] Failed preview on folder should be dismissed (if 4 images -> show 3) https://github.com/nextcloud/photos/pull/112
- [ ] Image load flash on preview (this is because [we use a trick to load the Image asynchronously to allow for request cancellation](https://github.com/nextcloud/photos/blob/766c3af0708ee1ebac590b8f235b07c891c67a32/src/components/File.vue#L108) if the user browse away, otherwise it will wait until all pictures are loaded and then navigate away
- [x] Navigation should ellipsis the folder name (for you @jakobroehrl ?)
- [ ] Opening the sidebar resize the files previews, we should reflow the list
- [x] Locations should open in a new tab

@skjnldsv let me know what you think and if I’m missing anything :)

jancborchardt:
@skjnldsv I think most of the list items in the original comment are good first issues for @jakobroehrl? :)

skjnldsv:
- [ ] Failed preview on folder should be dismissed (if 4 images -> show 3)
[...]

PR #29, opened 2019-12-06
Show navigation on empty folder too

skjnldsv:
see #6


jakobroehrl:
Empty folders aren't shown in the Album view, how can I test this?

skjnldsv:
You can enter an empty folder manually :)
Create one in files and use it in the url

PR #53, opened 2019-12-18
css for ellipsis

jakobroehrl:
Signed-off-by: Jakob <jakob.roehrl@web.de>
Before:
grafik

After:
grafik

@skjnldsv 
The ellipsis is only working with width. But a fixed width doesn't make sense, how to fix this?
grafik


skjnldsv:
you need a max-width of 100% or something :)

jakobroehrl:
> 
> you need a `max-width` of 100% or something :)

Thanks, tested it before, it's not working. ;(

Mikescops:
I went for that fix 
image

skjnldsv:
code looks good

PR #71, opened 2019-12-26
(Bugfix) Reset Timeline before mount

### LINKED ISSUE #70: Photos get duplicate when going from "Your albums" to "Your photos"
All photos get duplicate when going from "your albums" to "your photos" tabs.

Initial situation:
image

Then after going to your albums and then go back to your photos:
image

Another iteration:
image

Mikescops:
Fix #70 

This is temporary fix, to have a better flow. We need to :

1. Count loaded items in timeline
1. Conditional request the fetchContent method


skjnldsv:
> 1. Count loaded items in timeline
> 2. Conditional request the fetchContent method

Could you elaborate? :)

Mikescops:
@skjnldsv So if I understood well the code.
Every loaded (images) files for the timeline are stored in a timeline array (also called this.$store in Timeline.vue).
What happen is that when you switch between another tab and "Your timeline", the array timeline is not cleared and we call fetchContent() again and it ends up duplicating the content.
The current fix I suggest with this PR is to reset the timeline. But for end user it will mean going back to the top of the timeline and load images again. If we want to let the users go back to the previous state he was we need to whether count the number of already loaded images OR page number OR if the timeline is already populated and call conditionally the fetchContent() method during the beforeMount()

Let me know if I'm clear enough!

skjnldsv:
> If we want to let the users go back to the previous state he was we need to whether count the number of already loaded images OR page number OR if the timeline is already populated and call conditionally the fetchContent() method during the beforeMount()

Ah, well I'm not sure it will work anyway. Does the browser understand the scroll position? I think having a scroll-to-top button and remembering the position would be to discuss for a followup version :)

I think your patch is good! :+1:

skjnldsv:
@Mikescops I added you to our team, this should allow you to not create forks and work here directly, that will be easier :)

Mikescops:
I think it will be able when using "Go back" button, and to be honest having the timeline array already populated could save a huge amount of calls/time too!
But that also means any modification will be cached so it won't.
For sure it could happen in a future iteration.


I'm lock with the CI and this js folder, not sure how to fix it.

Ok thanks a lot! I'll keep it that way for the current PR.

Mikescops:
Ok I'm done with it :+1:

PR #77, opened 2019-12-26
(Fix) Reduce size of folder title

Mikescops:
Small fix for the folder title to be 16px as requested in #6


skjnldsv:
See https://github.com/nextcloud/photos/issues/6#issuecomment-555059516 too @jancborchardt :)

PR #78, opened 2019-12-26
(Bugfix) Resizing grid wasn't updating rows number

### LINKED ISSUE #73: Rows make photos disappear when resizing window
I'd like to challenge the use of rows in the grid. 

### Issue

The current way splits array of photos into multiple chunks of 5 photos. This leads to mistakes when the view/grid gets resized and can't display this amount of photos.

For instance here is a demo:
image

8 photos are currently displayed here.
And then when we resize:
image

6 photos being displayed on mobile view.

### Why current use of rows breaks in the future:

It would be great to have photos appearing without thumbnailing, in their original ratio aspect. That means the row will have a variable number of photos in it.

Also, if you take an upcoming feature to split by date the grid, this will endup in lot of Array slicing operations that shoud be handled by css styling.

Mikescops:
This PR adds an eventListener on the VirtualGrid to catch window resizing.

It fixes #73.


Mikescops:
This one needs a bit of QA with some larger dataset of images.

skjnldsv:
Needs rebase :)

PR #90, opened 2020-01-07
(Bugfix) Encode url for albums

### LINKED ISSUE #89: Wrong URL link when using special characters in folder title
When naming an album with a question mark for instance "Question ?", the generated url is wrong and the user is redirected to an empty folder.

image

The "%3F" is missing in the generated url:
image

@skjnldsv is the automatic redirection when visiting an empty album intended?

Mikescops:
Try to fix #89. 
I try to encode the url at the folder level, but maybe we need to make this earlier in the "pipeline".


Mikescops:
@skjnldsv I think there is a bigger issue here. We shouldn't have the url path using a user folder name. 

Here is what the file app do:
`apps/files/?dir=/Documents/question %3F&fileid=239`

Here is what the photo app do:
`apps/photos/albums/Documents/question ?`

2 issues:
1. The path is not encoded that can lead to injections in url 
1. Also the question mark `?` is understood as the query parameters separator

skjnldsv:
> * The path is not encoded that can lead to injections in url

We do not handle the path other than as a string, so I don't think you can inject anything with it. Worse case, you end up with a wrongly formatted url and the server files listing request will fail. Were you able to generate something?


> 2\. Also the question mark `?` is understood as the query parameters separator

I'm fine with encoding everything aside slashes :)
I feel like it's nice to have a visually pleasant url

What do you think? Wan we work around this?

Mikescops:
Problem is that the question mark encoded or not will be understood by the browser as the start of query parameters.
So when providing `"apps/photos/albums/Documents/question ?"`, nextcloud app will get `Documents/question%20` for the path.

Mikescops:
For me we are forced to do it that way `?dir=/Documents/question %3F` using a query parameter

skjnldsv:
> So when providing `"apps/photos/albums/Documents/question ?"`, nextcloud app will get `Documents/question%20` for the path.

looks good to me?

Mikescops:
> > So when providing `"apps/photos/albums/Documents/question ?"`, nextcloud app will get `Documents/question%20` for the path.
> 
> looks good to me?

Nope we are missing the question mark, then nextcloud cannot find the folder. Am i clear?

skjnldsv:
> Nope we are missing the question mark, then nextcloud cannot find the folder. Am i clear?

Right, I misread your comment.
What about https://dev.test.com/apps/photos/Test%20%3F ?

Mikescops:
Not working the browser interpret the %3F as ? on the fly

skjnldsv:
> Not working the browser interpret the %3F as ? on the fly

For me it's working :)
Capture d’écran_2020-01-07_17-13-20

skjnldsv:
But if it's too much work, let's switch back to using ?path=

Mikescops:
When you direct access or when you click on a href link?

skjnldsv:
> When you direct access or when you click on a href link?

direct

Mikescops:
Yep when using a <a> tag the browser unencode %3F

Mikescops:
Ok I found a fix, that's the best i can come with, you may want to find a better solution tho'

Mikescops:
Maybe we could unit test this :+1:

skjnldsv:
> Maybe we could unit test this

I dream of having the time to do this :p

src/utils/fileUtils.js
```diff
@@ -118,6 +118,10 @@ const genFileInfo = function(obj) {
 					: data
 			}
 		}
+
+		if (key === 'filename') {
+			fileInfo[camelcase(key)] = escape(data)
```
skjnldsv: We should avoid touching the fileinfos. They should stay as pristine as what the dav backend is giving us :)
I would rather have this fixed in  the vue router side: https://github.com/nextcloud/photos/blob/7ed958ced085f4548dafe7646d26e801e39732e4/src/router/index.js#L57

Mikescops: I'll try

rullzer:
Can't we do it the stupid approach..

`..../FILEID/PATH`.

Then the code can use the fileid as the actual location. But still have the path as fancy pancy that you see where you are...

src/utils/fileUtils.js
Mikescops: Can't find a way to do this..

skjnldsv: Let's talk tomorrow :)
You're here already?

Mikescops: I'll be there on Friday :/

skjnldsv: Oh, you're only there for two days?

Mikescops: Yes, I'll explain you why, but we're underwater at work

skjnldsv: Arf :(
I'll figure it out :)

PR #91, opened 2020-01-07
Fix viewer list

### LINKED ISSUE #72: Viewer slideshow iterate over folder
When launching the viewer slideshow it will iterate over the files in the same directory as the opened file.
This is an issue when using timeline for instance as we would like to iterate over all opened files.

skjnldsv:
Fix #72 

Needs latest viewer master (https://github.com/nextcloud/viewer/pull/335)


Mikescops:
Just reopening this, what if the list is not fully loading in the grid? can we style access the next images? (i guess no)

skjnldsv:
What do you mean by fully loaded?

Mikescops:
I mean if there are photos in not yet loaded rows, they won't be part of the list right?

skjnldsv:
We need to update the list once it changes indeed.
But rows are not loaded for the virtual grid, but the full list still exists (so it's just a matter of pagination)
I'll take care of it in a followup

skjnldsv:
Ah right, you cannot scroll while the viewer is opened, so it will not fetch the following images! Nice catch @Mikescops :+1: 
I'll think of it!

Mikescops:
That's what i meant, you're welcome!

PR #113, opened 2020-01-14
Fix shrinking of icons in long folder name

skjnldsv:
1. Narrow window with long folder
2. See icon get smaller than 44px width


src/components/Navigation.vue
```diff
@@ -191,6 +196,7 @@ $previous: 0;
 			}
 			&--root &__title {
 				padding-left: #{$marginW}px;
+				padding-left: #{$marginW}px;
```
Mikescops: Duplicate here?

skjnldsv:
```suggestion
```

skjnldsv: fixed

PR #119, opened 2020-01-15
(Bugfix  #115) Loading spinner position not centered

### LINKED ISSUE #115: Loading spinner doesn’t show centered in the view
Instead of showing centered in the content view, the loading spinner is half obscured by the header bar. This is likely because the container of the spinner is not very tall → should stretch to viewport height.

Mikescops:
Fix #115. Add more spacing on top of the spinner.


src/Photos.vue
```diff
@@ -82,6 +82,10 @@ export default {
 	flex-grow: 1;
 	flex-direction: column;
 	align-content: space-between;
+
+	&.icon-loading::after {
+		margin-top: 16px;
```
skjnldsv: It's probably safer to make sure the container have a min height of 100%? :)

Mikescops: What you mean ?

skjnldsv: That the icon loading is centered on the parent space.
The real issue is that the parent (the app container) should have a minimum height of 100%

Mikescops: Oh got it!

Issue #142, opened 2020-01-21
Allow to browse the folder tree


atenart:
Hello,

I think we could benefit from having an option (like adding a category on the left) to allow users to browse their photos directly using the folder tree of their data. I understand having an intelligent sorting and abstracting the actual locations of photos is nice for some, but there are cases were accessing the real files tree is more usable; for example when having a shared directory between nextcloud and other apps, which is manually or automatically sorted by year / month… In such case I end up with far too many albums, of which many are named the same: it's difficult to know which one I should open, and it's also not easy to find the album in the (huge) list.

One other option could be to allow to launch the photo app from a directory in the files browsing.

Thanks!

skjnldsv:
@jancborchardt thoughts?

jancborchardt:
> One other option could be to allow to launch the photo app from a directory in the files browsing.

This cross-linking is key – that way we don’t have to duplicate stuff. In the 3-dot menu next to the title / share icon in the Albums view, we could have an entry saying "Open in Files".

nemihome:
It does not make sense to show folder names without a chance to rename them or showing the tree in album view. In my case I have a hierarchy like 
year - destination (first folder level) and then in every folder 3 subfolders gpx / info / pictures. That results on every folder has the name in the photos app because they are all subfolders with the same name That does not help and a tree view or a chance to use tag view where I can tag the folders would help a lot.

adi2k5:
There definitely should be an option to browse gallery with folder tree like it was in old Gallery. I just migrated to NC18, and realized, that browsing my albums now is a tragedy. I have several hudreds of photo folders from last over 15 years, and on top of that, new Gallery, beside all folders with photos, picked also another several dozens of folders from other folders than my photos (mostly folders with some icons etc). This way, when I'm going now to Album view, I'm immediately killing my browser tab at least for several minutes, sometimes killing it completely. This is happening most likely because of generating previews, even with preview generator enabled, From last ~10 tries to view any albums, only two ended up with success after waiting several minutes, for others I had to restart tab (firefox) or whole tab group with nextcloud (vivaldi) as they simply crashed. Thus option to browse folder structure and include/exclude locations proposed in some other thread would be really very welcomed for bigger galleries like mine (almost 2 TB of photos).

skjnldsv:
> with folder tree like it was in old Gallery

was it though? I don't think you could, you were only able to browse folders that had images, and this is what we do :)

adi2k5:
[...]

PR #159, opened 2020-01-27
Respect .noimage and .nomedia files

### LINKED ISSUE #75: Not taking into account the .noimage flag
Folders with the file `.noimage` are still viewable.

rullzer:
Fixes #75

When a folder has a .noimage or .nomedia node. Just don't show the
folder in the album overview. Ignore it and do not traverse it.

Signed-off-by: Roeland Jago Douma <roeland@famdouma.nl>

### DIFF
--- a/lib/Controller/AlbumsController.php
+++ b/lib/Controller/AlbumsController.php
@@ -142,6 +142,11 @@ class AlbumsController extends Controller {
 			return [];
 		}
 
+		// Ignore folder with a .noimage or .nomedia node
+		if ($folder->nodeExists('.noimage') || $folder->nodeExists('.nomedia')) {
+			return [];
+		}
+
 		$nodes = $folder->getDirectoryListing();
 
 		foreach ($nodes as $node) {


skjnldsv:
Of course it still won't work for the timeline, I wonder if I can do it with dav search :thinking:

rullzer:
> Of course it still won't work for the timeline, I wonder if I can do it with dav search thinking

Not really.

Mikescops:
@skjnldsv if we add the ability to select the folders for the timeline, it will be fine

skjnldsv:
> @skjnldsv if we add the ability to select the folders for the timeline, it will be fine

I guess it's a nice workaround yes :)

newhinton:
@skjnldsv is this feature for the timeline planned? I tried 1.1.0 and for albums it works flawles, but not for the timeline (as expected). But honestly, the timeline should also check for that. Is there anything i could do? Also, is there a seperate issue for the timeline?

skjnldsv:
@newhinton that is a good question!
Is is technically possible, no idea :)
There is also #142 which is for the timeline.

But I'm not sure I can exclude path containing nomedia/noimages while doing a dav SEARCH. Any idea?
I could search for all nomedia/noimages and then exclude those paths from the main SEARCH I guess :)
Not sure how it scale though :thinking:

PR #186, opened 2020-02-11
Avoid passing an incomplete cancel token object to axios

juliushaertl:
I had the following error thrown on my development instance when running photos from the master branch:

Timeline.vue:194 Error fetching timeline TypeError: t.cancelToken.throwIfRequested is not a function
    at s (dispatchRequest.js:13)
    at t.exports (dispatchRequest.js:24)

Strangely when bisecting this, it only occurs after 8d715bc (Bump webdav from 2.10.1 to 2.10.2) but I cannot explain that when looking at the diff from the webdav library.

I looked a bit deeper and it seems that when passing the cancel token to axios it somehow only has a promise property while axios expects the throwIfRequested to be present.
image

I'm not sure if that fix actually makes 100% sense but it fixes the issue for me. :wink:


Mikescops:
Do you mind merging this one?

juliushaertl:
I'd like to have @skjnldsv to have a look before merging. :wink:

skjnldsv:
Hey :)
I would rather have it fixed here: https://github.com/nextcloud/photos/blob/88ed0ab4e288a56abc2e8b60536826ad1ea194dc/src/views/Timeline.vue#L95
https://github.com/nextcloud/photos/blob/88ed0ab4e288a56abc2e8b60536826ad1ea194dc/src/views/Timeline.vue#L150

This is because we created a fake cancel request, but instead of always running it, we should check against null and exec accordingly :)

skjnldsv:
also: https://medium.com/@decker/typeerror-config-canceltoken-throwifrequested-is-not-a-function-6b7b20b54fa2

src/patchedRequest.js
```diff
@@ -31,7 +31,9 @@ const oldPrepareRequestOptions = request.prepareRequestOptions
 request.prepareRequestOptions = function(requestOptions, methodOptions) {
 	// add our cancelToken support
 	if (methodOptions.cancelToken && typeof methodOptions.cancelToken === 'object') {
-		requestOptions.cancelToken = Object.assign({}, requestOptions.cancelToken || {}, methodOptions.cancelToken)
```
skjnldsv: So in fact, we should always use the given methodOptions.cancelToken instead of requestOptions.cancelToken.
So let's just assign it here and we'll be able to keep all the prototyping now

PR #212, opened 2020-02-28
Don't flatten out albums

### LINKED ISSUE #197: No folder hierarchy in the "Your albums" tab is confusing
Hello,

Version used: NC 18.0.1 upgraded from NC17; Using LDAP backend.

Within our org, we have a shared media folder. The folder is divided into subfolders by date like this:

```
|-- 2018
|   |-- event1
|   |-- event2
|   `-- event3
|-- 2019
|   |-- event1
|   |-- event2
|   `-- event3
`-- 2020
    |-- event1
    |-- event2
    `-- event3
```

This goes back years and contains hundreds of event folders.

The problem is, that when clicking on "Your albums" in the new photos app, all folders and subfolders show up at the same time in a completely flat hierarchy, so we see hundreds of "event1, event1" etc. folders without any context. Sorting by folder name is also not possible.

I expect there to be a hierarchy, similar to the previous gallery app, which goes by the folder hierarchy. Or perhaps an additional tab "Your Folders" and "Shared Folders" where the folder hierarchy is preserved.

Thanks a ton & happy to provide more debug details.

Related issue: #120

rullzer:
Before:
before

After:
after

Fix #197

### DIFF
--- a/lib/Controller/AlbumsController.php
+++ b/lib/Controller/AlbumsController.php
@@ -116,8 +116,8 @@ class AlbumsController extends Controller {
 		yield $folder;
 
 		foreach ($nodes as $node) {
-			if ($node instanceof Folder) {
-				yield from $this->scanFolder($node, 0, $shared);
+			if ($node instanceof Folder && $this->scanFolder($node, 0, $shared)) {
+				yield $node;
 			} elseif ($node instanceof File) {
 				if ($this->validFile($node, $shared)) {
 					yield $node;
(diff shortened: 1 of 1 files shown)


skjnldsv:
So it still display the full path if it find a picture in like any of the subfolder, right?

rullzer:
> So it still display the full path if it find a picture in like any of the subfolder, right?

basically instead of only adding a folder that contains pictures. It now adds the top folder if it contains a subfolder with pictures.

PR #343, opened 2020-05-28
Hide the location entry if Maps is disabled and user is not an admin

georgehrke:
To reproduce:
- Disable Maps app
- Login as non-admin user
- Click the "Location" menu entry:

It will show you this error message:
F7CC82A7-8F94-4827-B034-2852D480DB71

Only admins can access this page.

Solution:
- Do not display the "Location" menu entry if Maps is disabled and the current user is no admin.


juliushaertl:
Bundles seem not up to date :wink:

georgehrke:
> Bundles seem not up to date 😉

No idea where that error is coming from. This is branched off directly from master and i just committed all the files `make` changed.

src/Photos.vue
```diff
@@ -71,6 +75,7 @@ export default {
 			svgplaceholder,
 			imgplaceholder,
 			videoplaceholder,
+			showLocationMenuEntry: OC.isUserAdmin() || isMapsInstalled,
```
ChristophWurst, line 78: why not use https://nextcloud.github.io/nextcloud-auth/modules/_user_.html#getcurrentuser, though? :)

georgehrke: Excellent idea ;)

georgehrke:
The changes to the actual package.json got lost. Let me take care of that.

Issue #277, opened 2020-04-07
Filter to show only Videos


ghost:
**Is your feature request related to a problem please describe**
Can we please have a filter to show videos only?

**Describe the solution you'd like**
A simple filter tick box or select on the left hand side to show only videos?

Or maybe under “Your Photos” have “Your Videos”?

skjnldsv:
> Or maybe under “Your Photos” have “Your Videos”?

I like the idea! @jancborchardt  ?

PR #392, opened 2020-07-15
show video filter

jakobroehrl:
https://github.com/nextcloud/photos/issues/277

grafik
grafik
grafik


src/views/Timeline.vue
```diff
@@ -124,6 +132,22 @@ export default {
 			// reset component
 			this.resetState()
 
+			// content is completely different
+			this.$emit('update:loading', true)
+			this.fetchContent()
+		},
+		async onlyPhotos() {
```
skjnldsv, line 139: You can't do this, it will create issues as `this.onlyPhotos` is both a method and a prop ?

skjnldsv: Ah, sorry I misread. I thought this was in methods, not in watch! All good! :100:

src/views/Timeline.vue
```diff
@@ -124,6 +132,22 @@ export default {
...
+			// reset component
+			this.resetState()
+
+			// content is completely different
+			this.$emit('update:loading', true)
+			this.fetchContent()
+		},
```
skjnldsv, line 146: We could maybe merge the three duplicate functions into one?

jancborchardt:
With having a specific video view, the issue of the videos not having previews becomes a much bigger problem.

Is it possible to get any kind of preview at all out of the videos here? Otherwise a specific video view doesn’t seem that much of an enhancement, if it’s just all loading filetype icons and then you need to get lucky to find the correct video. ;)

(Actually it makes it a less good UX since it pushes a shortcoming of Photos into the forefront.)

jakobroehrl:
> 
> 
> With having a specific video view, the issue of the videos not having previews becomes a much bigger problem.
> 
> Is it possible to get any kind of preview at all out of the videos here? Otherwise a specific video view doesn’t seem that much of an enhancement, if it’s just all loading filetype icons and then you need to get lucky to find the correct video. ;)
> 
> (Actually it makes it a less good UX since it pushes a shortcoming of Photos into the forefront.)

On my productive system I have no problems with missing video previews. It's only on the dev system.

skjnldsv:
> Is it possible to get any kind of preview at all out of the videos here?

Nope, it all depends on your setup. If you have video previews properly setup :/
https://www.allerstorfer.at/nextcloud-install-preview-generator/

jakobroehrl:
Todo: 
We want to add a notice on the video pane if not installed:

> consider installing the app preview generator for showing video previews here

jancborchardt:
@jakobroehrl rather, I would say to only show the split "Your photos" and "Your videos" view when that preview generator is installed and configured.

If it is not, then the videos view is not so useful and we only need the Timeline, and the notice about previews can be shown to admins either:
- On top of the pictures
- In the bottom left settings area
- On the bottom part of the navigation

tacruc:
Note: The video preview might be disabled even if the preview generator app is enabled. 
I had quite a fight today getting it to work, even reading all the docs. As everything, which was mentioned in all docs was installed, but still video preview was not available. Therefore I think the hint of just install this one app might lead to more frustration, than it helps. 

The underling problem was, that most video codecs are not opensource and therefore not shipped with many distributions. Even installing ffmpeg is not sufficient, if the codecs are missing. 
But how many and which codecs have to be installed to consider video previews as enabled? Is webm sufficient? Or webm and H.263. 

In short I think it will be hard to check if video previews are available and working on a system.

skjnldsv:
> In short I think it will be hard to check if video previews are available and working on a system.

@rullzer is this feasible to create a setting that checks each provider and if it's properly working?
Maybe that would be cool to be able to narrow and debug things so we know what preview provider is enabled and which one of them is not working (I also have the same issue, video enabled, but not generating previews, though I did not debug further yet)

rullzer:
> > In short I think it will be hard to check if video previews are available and working on a system.
> 
> @rullzer is this feasible to create a setting that checks each provider and if it's properly working?
> Maybe that would be cool to be able to narrow and debug things so we know what preview provider is enabled and which one of them is not working (I also have the same issue, video enabled, but not generating previews, though I did not debug further yet)

Not really. You actually need to encode something and check if the result is correct. Which is almost impossible.

jakobroehrl:
Sad, but then let's close this PR

PR #398, opened 2020-07-24
Show groupfolders in shared folders

### LINKED ISSUE #129: Support Groupfolders
We use group folders app for shared space and most apps support them without any issue, but new photo app won't show any folders/images.

rullzer:
Fixes #129 

So. The code is simple. But I did notice it taking longer on my system. Also mainly because group folders then to be freaking huge ;)

CC: @xf- @gabrielklein @jakobroehrl 

But should be fine in the 99% case.

### DIFF
--- a/lib/Controller/AlbumsController.php
+++ b/lib/Controller/AlbumsController.php
@@ -99,7 +99,6 @@ class AlbumsController extends Controller {
 				'etag' => $node->getEtag(),
 				'fileid' => $node->getId(),
 				'filename' => $path,
-				'etag' => $node->getEtag(),
 				'lastmod' => $node->getMTime(),
 				'mime' => $node->getMimetype(),
 				'size' => $node->getSize(),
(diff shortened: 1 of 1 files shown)


lib/Controller/AlbumsController.php
```diff
@@ -136,7 +135,8 @@ private function validFile(File $file, bool $shared): bool {
 	}
 
 	private function isShared(Node $node): bool {
-		return $node->getStorage()->instanceOfStorage(SharedStorage::class);
+		return $node->getStorage()->instanceOfStorage(SharedStorage::class) ||
+			$node->getStorage()->instanceOfStorage(\OCA\GroupFolders\Mount\GroupFolderStorage::class);
```
skjnldsv, line 139: `use OCA\GroupFolders\Mount\GroupFolderStorage` ?

rullzer: No lets not import it with use. As we can't know for sure it is available.
This just expands to the classname.

Issue #417, opened 2020-08-14
show if it's a photo or a video


jakobroehrl:
It's not possible to recognize if it's a photo or a video file.
Maybe it would be nice to show a litte icon inside every video file.

PR #419, opened 2020-08-16
show if is's a pic or vid

jakobroehrl:
solves #417 

grafik

should work with video previews, too.


jakobroehrl:
@skjnldsv 
1. How could I solve this better then with this ActionButton? I only want to show an video/picure icon
2. How to show this icon over the picture? Currently the picture covers it.
Thanks for you help

skjnldsv:
Hey!
Since there is no action on the button, it would be easier with a simple div, absolutely positioned, with the icon-video class :)

jakobroehrl:
This is working, but it's covered if there are any previews:
grafik

src/components/File.vue
```diff
@@ -132,6 +133,12 @@ export default {
 <style lang="scss" scoped>
 @import '../mixins/FileFolder.scss';
 
+.icon-video {
+	position: absolute;
+	top: 10px;
+	right: 10px;
```
skjnldsv, line 139: Add a `z-index: 5` Or higher if 5 is not enough :)

jakobroehrl:
It's working nice, thanks.

jancborchardt (approved):

PR #435, opened 2020-09-01
Non-cropped layout

### LINKED ISSUE #143: No cropping photo design
In "Photos" all photos are cropped square. This representation already exists in the grid view. But in "Photos" the photos (and videos) should be in the foreground. They should therefore be presented uncut. At least there should be an option in "Photos" to present photos and videos uncropped and to share this view via public link.

Here are a few nice examples:
https://www.flickr.com/photos/dmtoh/
https://www.photocase.de/p/491889
https://stock.adobe.com/contributor/204990950/weerasak

Photos and videos are also presented uncut in Google Photos.

skjnldsv:
Capture d’écran_2020-09-01_11-03-12

Fix #143


Mikescops:
@jancborchardt @skjnldsv good to go

src/mixins/UserConfig.js
```diff
@@ -0,0 +1,64 @@
...
+import { emit, subscribe, unsubscribe } from '@nextcloud/event-bus'
+import { generateUrl } from '@nextcloud/router'
+import { loadState } from '@nextcloud/initial-state'
+import axios from '@nextcloud/axios'
+
+const eventName = 'photos:user-config-changed'
+
+export default {
+	data() {
+		const croppedLayoutLocalStorage = localStorage.getItem('photos:croppedLayout')
```
skjnldsv, line 32:
```suggestion
		const croppedLayoutLocalStorage = localStorage.getItem('photos:cropped-layout')
```
Funny we had a similar discussion today on another repo :p
https://github.com/nextcloud/nextcloud-vue/pull/1457#discussion_r505550454

skjnldsv: Ah, it's because we use `localStorage.setItem('photos:' + setting, value)` later, right.
Let's keep it like that, it's no big deal. It's not like it is a `nextcloud:xxx` one :+1:

Mikescops: I think it's cool to have the name of the app only, we know we're on nextcloud namespace anyway

skjnldsv: > I think it's cool to have the name of the app only, we know we're on nextcloud namespace anyway

No, that's not it. nextcloud: would not be fitting here anyway. This is for global events/variables only.
I was referring to the kebab/camel case :)

skjnldsv:
Can't approve my own pr, but looks good :+1:  :)

Mikescops (approved):
Can approve my half PR

Issue #120, opened 2020-01-16
Add albums feature and separate "folders" from "albums"


ma12-co:
At the moment pictures need to physically be into the same folder to form a shareable collection. In my opinion, the current 'albums' navigation item should be called 'folders' as it provides a nicer way to browse photos in the folder structure, but aside from that It would be nice to be able to create any number of shareable collections of pictures (albums) without having to duplicate them in different folders.

This is how the menu would look like: 

* photos
* folders
* albums
* shared albums
* tagged photos
* location

Mikescops:
It's a good idea (maybe Timeline instead of Photos) but this is related to #6 and #84 too.

ma12-co:
cc @nextcloud/designers

kaffeeringe:
I miss the old handling of folders. It was everythiny but perfect, but at least I was able to navigate it. 
Shotwell sorts all photos into folders of years, months and days. In Nextcloud it's now just 1 gazillion folders with numbers as names. And too many for previews to render...

jancborchardt:
> It would be nice to be able to create any number of shareable collections of pictures (albums) without having to duplicate them in different folders.

Absolutely agree. It should be possible to select pictures and add them to "virtual" albums. We need to get a bit away from the filesystem paradigm here.

However, before we do this we should make sure we can integrate this with the mobile apps for feature parity, cc @tobiasKaminsky @marinofaggiana @AndyScherzinger.

tobiasKaminsky:
This needs some real deep thinking, adding a "virtual" structure on top of existing folder structure is a big challenge, both UX and tech-wise.

skjnldsv:
Yep :/
If we go for this, better drop the folder structure entirely and suggest marking all folder content as a single album instead. Also, we'd lose compatibility with other photos app I assume? Android is folderbased, no idea for iPhotos. But I ideally, opening your nextcloud data in iPhotos should also display your albums.

kaffeeringe:
Please keep in mind that people already have hundreds of Photos in their Nextcloud.
The normal folders are not a replacement, because folders don't have a preview there. It's pretty hard to find certain photos within a complex folder structure.

tobiasKaminsky:
To not make this too complicated, I suggest to first find a good way within this photo app.
Then it should be somehow integrated into NC server, and then we can think about integration into clients.

For now I always understood the photo app as a different view onto photos in NC, with the same hierarchy of folders.
[...]

PR #457, opened 2020-09-22
Rename "albums" to "folders"

strugee:
This more accurately reflects what the feature actually is, and paves the way for the addition of a "real" virtual albums feature as discussed in #120.

This is my first significant contribution to any Nextcloud app, so when reviewing, please do not assume I know what I'm doing :P


strugee:
CI is failing because the Webpack build leaves the working tree dirty. I'll push an amended patchset that includes the changes to them, since AFAICT that's what I'm supposed to do? The contributing docs don't mention it (I'm just guessing based on the nextcloud/server README).

strugee:
Oh, another note: I did not delete localization strings for "Your album" under the assumption that they will be needed soon. (Indeed locally I'm working through adding a virtual albums feature that would use them again.)

jancborchardt:
Sorry for the late reply @strugee!

I’m not entirely sure about this. We should **not** have both "Folders" and "Albums" as that will create confusion. Every folder should just be an album (as now, for fallback) and then the new possible Albums implementation should also allow virtual albums.

Or what do you think @skjnldsv @karlitschek @jakobroehrl @Mikescops @dassio @nextcloud/photos – since it’s more than just a wording change.

da-anda:
since you can not configure what is scanned/displayed in the photos app as "albums", I very much support that the current implementation is renamed to just "folders". The cover art of my music collection certainly is not an album and thus should never ever show up as such. Folders are folders and albums are a virtual collection of files from a single or multiple folders.

Mikescops:
To be honest @jancborchardt I'm fine with the Folders rename, it's the truth we just browse the file directory.

We shouldn't rely on what things could be in the future because as we know future features can be really far from now. And the current "Album" naming is indeed misleading.

karlitschek:
> Sorry for the late reply @strugee!
> 
> I’m not entirely sure about this. We should **not** have both "Folders" and "Albums" as that will create confusion. Every folder should just be an album (as now, for fallback) and then the new possible Albums implementation should also allow virtual albums.
> 
> Or what do you think @skjnldsv @karlitschek @jakobroehrl @Mikescops @dassio @nextcloud/photos – since it’s more than just a wording change.

No strong opinions here to be honest.

Mikescops:
Can we move on here?

strugee:
Meant to reply to this and forgot, sorry. I do like @jancborchardt's proposed design (we could call folders "Folder Albums" or something and have an option to hide them, for cases like @da-anda's music cover art) but realistically that feature might be a ways off.

I propose we merge this as-is since it more accurately reflects the current feature set. Whoever implements real virtual albums can just revert the name change. If that sounds agreeable to everybody I'll go ahead and rebase this.

da-anda:
In order to be more clear that the nodes do only list folders containing pictures, I think the shown headlines in the `my folders` and `my shared folders` nodes could be a bit clearer about this and say something like `Your picture folders`, `Folders containing pictures` or something like that.

jancborchardt (approved):
Let’s go with the rename then since it indeed reflects the current state. :) We can always adjust later on.

(Just don’t want the navigation to become filled with different terms of "Album" and "Folder" in the long run, to make that clear.)

skjnldsv:
@strugee please rebase (you can drop changes for `/js`)? :)

strugee:
@skjnldsv what do you mean drop changes for `js/`? The original patchset in this PR didn't include changes in `js/`, but CI complained about that so I added them. Won't that happen again?

@da-anda I'm on the fence; I see what you're saying but I think those suggestions would be too wordy for the UI. And it's the Photos app anyway, so I think there's probably enough context for the user.

skjnldsv:
> @skjnldsv what do you mean drop changes for `js/`? The original patchset in this PR didn't include changes in `js/`, but CI complained about that so I added them. Won't that happen again?

For the rebase, you can ignore all conflicts for js files as you will have to recompile them afterwards :)

strugee:
This should be ready for merge.

Issue #426, opened 2020-08-23
Showing dates/years on timeline scrollbar


dlauper:
**Is your feature request related to a problem? Please describe.**
My photo folder contains 43688 pictures. Going back more than a couple of hundred pictures is not really doable, since it would take a day to go back that far. 

**Describe the solution you'd like**
I propose a real timeline as or next to the scrolling bar. Kind of like [Plex](https://www.plex.tv/wp-content/uploads/2018/05/image-photos-videos-timeline-1440x1320.jpg)  but using the full height for spreading out the dates or like Google Photos and probably some others as well,

Cherry on top: if you drag the right-hand side vertical scrollbar to the left, it zooms in allowing for a more precise positioning. Or maybe when clicking it holding ctrl it zooms in to the range of a year. 

**Describe alternatives you've considered**
Album view. Which makes me never ever use the timeline. 

**Additional context**
I'm aware that this could be challenging given the way pictures are loaded step by step. I assume that somehow the total length of the page would need to be known beforhand. The upside is, that this concept could be applied to everything, not just dates, but alphabetical and size sorting too. 

This feature request is part of the big picture of my Photos vision discussion at the [forum](https://help.nextcloud.com/t/ideas-for-revolutionary-photo-organization/89278).

skjnldsv:
Similar: #45

dlauper:
> 
> 
> Similar: #45

Kind of, but #45 doesn't allow you to to go back to February 2008 for instance.

mattis75:
I value the photo app and would also appreciate a possibility to view and scroll between dates in the slideshow view. This would be great and make the app far more usable. regards. matthias

Mikescops:
This is a very complex feature given the Dav backend, we would have to pre-process all existing image in the background to create "groups" of photos by date to be able to know how much data we have per day/month/year.

skjnldsv:
> This is a very complex feature given the Dav backend, we would have to pre-process all existing image in the background to create "groups" of photos by date to be able to know how much data we have per day/month/year.

Unless we take just pre-defined ranges. Like for the last year we take every month and then every year
Then if you click we just show the closest :shrug:

Mikescops:
Could be completely irelevant if we don't have an idea of what the actual data is.

da-anda:
do frontend applications have to use the DAV backends? Just curious if there is some rule for this (I am new to NC)

skjnldsv:
[...]

PR #468, opened 2020-10-07
Implement new Vue Virtual Grid

Mikescops:
This implement the new virtual grid lib I created: https://www.npmjs.com/package/vue-virtual-grid

Also fast forward #125

And finally fix a bug with title rendering and the new sidebar.

Please take time to QA this to make sure there is no regression I couldn't catch.

Preview of it:

image

and:

image


src/components/SeparatorVirtualGrid.vue
```diff
@@ -0,0 +1,52 @@
...
+export default {
+	name: 'SeparatorVirtualGrid',
+
+	inheritAttrs: false,
+
+	props: {
+		item: {
+			type: Object,
+			required: true,
+			injected: {
```
skjnldsv, line 18: Does that exists? Never saw it before: `injected`
https://vuejs.org/v2/api/#props

src/components/FileVirtualGrid.vue
```diff
@@ -0,0 +1,143 @@
+
+<template>
+	<a :class="{'file--clear': !loaded}"
```
skjnldsv, line 3: Why do we need a new dedicated template?
Ideally we should keep using the same File component that any other Views. That way any change is visible and propagated to all! :thinking:

src/views/Albums.vue
```diff
@@ -30,30 +30,31 @@
...
-				:key="dir.fileid"
-				v-bind="dir"
-				:show-shared="showShared" />
-			<File v-for="file in fileList"
-				:key="file.fileid"
-				:list="fileList"
-				v-bind="file" />
-		</template>
-	</Grid>
+		<Grid>
```
skjnldsv, line 39: I feel like this is not related to this pr, no?
What is this trying to fix?

src/components/SeparatorVirtualGrid.vue
Mikescops: it's a custom property from my lib

src/components/FileVirtualGrid.vue
Mikescops: Because props are passed from the lib and as I didn't want to change the other views Grid system i created a new one, but i can remove it for sure

src/views/Albums.vue
Mikescops: I changed rendering of navigation to fit the new Virtual Grid but it was impacting the other views so I had to make a quick fix there also.

If we consider to add the virtual grid everywhere then it's fine!

Mikescops (commented):
I found few bugs:
- (not related to this PR) XML parsing has an issue when there is no images
- when resizing the window the computation goes wrong for full width items (this has to be fixed in the virtual grid lib)
- the viewer is locked to folders once again (maybe we will also have to edit the virtual grid lib to manually trigger the next batch of data)

src/components/SeparatorVirtualGrid.vue
skjnldsv: But it's not passed. you cannot nest props in Vue.
If you needed to pass a default, then you need to use the default key, `injected` means nothing to vue here. :)

Mikescops: ah right, i think i'm mixing things up with Typescript, you're right i only need to have the root property and i can still access to item.injected.xxxx

Mikescops: This is fixed

Mikescops:
>     * when resizing the window the computation goes wrong for full width items (this has to be fixed in the virtual grid lib)
> 
>     * the viewer is locked to folders once again (maybe we will also have to edit the virtual grid lib to manually trigger the next batch of data)

Both issues are now fixed + the ones reported by @skjnldsv 

@jancborchardt if you can give a try

jancborchardt (changes requested):
Very nice! Only some small feedback:
- With the new date headings, the "Your photos" heading is not necessary anymore and can be cut :) But make sure there’s some nice whitespace on top of the first date heading
- The date headings could do with some more whitespace above them, let’s say double of what’s there now (so 72px)
- I get this strange preview glitch when scrolling (in Firefox) where already loaded pictures fall back to displaying only the filetype icon:
Photos scroll loading glitch

Mikescops:
I have the same glitches and I'm not sure why they are there, and it seems somehow that the old grid what already doing that but less visible.

I'll try to find why and fix the other spacing issues.

Mikescops:
@jancborchardt it's fixed :smile: 
thanks @skjnldsv

skjnldsv:
> it's fixed

Congrats! Was not that easy :p

src/components/FileVirtualGrid.vue
```diff
@@ -0,0 +1,124 @@
...
+			})
+		},
+
+		onLoad() {
+			this.loaded = true
+		},
+
+		filterOnlyViewableFiles(list) {
+			return list.filter(item => item.injected.filename).map(item => item.injected)
+		},
```
skjnldsv, line 91: I don't understand the purpose of this function?
Why filtering the filenames? 
Couldnt this directly be managed by the loadMore function?

Mikescops: for the loadmore it's not necessary, true, but for the current list it is because i'm injecting title/separators

Mikescops: i was wrong it's necessary for both

src/views/Timeline.vue
```diff
@@ -143,47 +123,69 @@ export default {
...
 			const { request, cancel } = cancelableRequest(getPhotos)
 			this.cancelRequest = cancel
 
 			try {
-				// get content and current folder info
 				const files = await request(this.onlyFavorites, {
-					page: this.page,
-					perPage: this.gridConfig.count * 5, // we load 5 rows,
+					page: params.offset,
+					perPage: 30,
```
skjnldsv, line 143: Use the grid config

src/views/Timeline.vue
```diff
@@ -205,63 +211,29 @@ export default {
...
 		},
 
 	},
 
 }
 </script>
+
+<style lang="scss" scoped>
+.grid-container {
+	padding: 66px;
```
skjnldsv, line 237: Padding is variable.
Please check https://github.com/nextcloud/photos/blob/b501e72f31bbaf90d89e3c55ef2438b002244d0b/src/assets/grid-sizes.js

src/components/SeparatorVirtualGrid.vue
skjnldsv, line 18: Still here

src/components/SeparatorVirtualGrid.vue
```diff
@@ -0,0 +1,52 @@
...
+				type: Number,
+				required: true,
+			},
+		},
+	},
+}
+</script>
+
+<style lang="scss" scoped>
+.grid-title {
```
skjnldsv, line 40: Grid title should be using the grid config sizing (padding, margins...)

src/components/SeparatorVirtualGrid.vue
skjnldsv, line 18: `injected` needs to go away.

src/views/Tags.vue
```diff
@@ -27,37 +27,38 @@
...
-				</template>
-			</EmptyContent>
-
-			<File v-for="file in fileList"
-				:key="file.fileid"
-				:list="fileList"
-				v-bind="file" />
-		</template>
-	</Grid>
+		<Grid>
```
skjnldsv, line 36: I don't see how this is relevant to this pull request?

src/views/Timeline.vue
skjnldsv, line 143: Grid config is not used anymore

src/views/Timeline.vue
```diff
@@ -143,47 +123,69 @@ export default {
...
-				// next time we load this script, we load the next page if the list returned
-				if (files.length === this.gridConfig.count * 5) {
-					this.page++
-				} else {
-					console.debug('We loaded the last page')
+				if (files.length === params.offset === 0) {
+					this.isEmpty = true
+				}
+
+				if (files.length !== 30) {
```
skjnldsv, line 150: Same here (grid config)

src/views/Timeline.vue
```diff
@@ -143,47 +123,69 @@ export default {
...
+							list: this.$refs.virtualgrid.getCurrentItems,
+							loadMore: this.$refs.virtualgrid.loadMoreData,
+						},
+						width: 256,
+						height: 256,
+						columnSpan: 1,
+						renderComponent: FileVirtualGrid,
+					})
+					return finalArray
+				})
```
skjnldsv, line 188: I would love a little bit more comments here, so anyone can follow :see_no_evil: 
I prefer to have more comments than needed so at least it's clear :v:

skjnldsv (changes requested):
- [x] Header Navigation is  missing when there is pictures (present when there is none)
- [ ] Loading indicator is missing
- [ ] Grid sizes config is not used anymore
- See inline comments too :)

src/views/Timeline.vue
Mikescops: yes, it is useless has the grid is not based on row number anymore, we can decide (or let the user decide) of the number of images to load by batch.

src/views/Timeline.vue
Mikescops: Shall we make a gridconfig that is a setting?

src/views/Timeline.vue
Mikescops: Sure ok

src/components/SeparatorVirtualGrid.vue
Mikescops: We can't remove injected, it's part of the lib. Everything inside the item is edited by the lib, `injected` is set by the user of the lib, it can be anything and the lib won't touch it, in our case it's whether an image or a separator.
If you want to better understand check out the demo of the lib.

src/components/SeparatorVirtualGrid.vue
Mikescops: Yes understood

Mikescops:
> Header Navigation is missing when there is pictures (present when there is none)

@jancborchardt asked to remove it when there is pictures

> Loading indicator is missing

I'm not sure how we can re-add the loading indicator, the lib is not ready for this, I think we can open an issue on the lib and fix this later, I'm not convinced it's a blocker

> Grid sizes config is not used anymore
As answered in the small comments, gridconfig is now handled on the lib side. It is configurable through functions passed to the grid component. In my opinion we can remove it and set the batch size with a param.

src/views/Tags.vue
Mikescops: I updated the css of the navigation to fit the new grid, and it is impacting the other pages, so I removed the navigation from being inside the grid :)

src/views/Timeline.vue
Mikescops: reverted

src/views/Timeline.vue
Mikescops: based on gridconfig now

src/views/Timeline.vue
Mikescops: I added comments

src/components/SeparatorVirtualGrid.vue
Mikescops: Removed, i misunderstood the comment sorry

Mikescops:
I have fixed all suggested issues except the loading icon when waiting for next batch.
For this one I created an issue on the lib side: https://github.com/Mikescops/vue-virtual-grid/issues/8
If it's a non blocking regression we can move forward here.

src/views/Tags.vue
Mikescops: It is due to the new styles applied to titles to fit the new grid. I had to fix them on other pages.

skjnldsv (approved):

jancborchardt:
Nice! Only found little issues:
- [ ] `.grid-container` has a white background set which shows in dark mode. It doesn’t actually need any background set? (If it does then we should use variables – but I really don’t think the background needs to be there?)
- [ ] There’s a bit too much space on the top. It’s the padding-top of the heading _plus_ the 66px of padding on the grid-container. I’d just remove the 66px top on the grid container, for any form factor. The padding of the heading is enough :)
- [ ] The bottom of the photo list has very little space, I’d go for something like 256px down there to signify the list is over.
image

Mikescops:
@jancborchardt nice catch! done

jancborchardt (approved):
Looks good design-wise! :) Nice!

Mikescops:
@skjnldsv can I let you merge this?

TKFRvisionOfficial:
Hey guys, that is maybe a long shot but have you thought about implementing a little side timeline (a litle line with points on it on the site) where every point is a month so you can easily find pictures. Like at amazon cloud?

Mikescops:
See #426

sanchosk:
Would it be somehow possible to separate not only by months, but also by days?

PR #477, opened 2020-10-12
Implement "your videos" tab

Mikescops:
This is a fix for request #277.

It implements a new "Your videos" tab and the ability to filter a bit more mime types when requesting photos.

image

I'd like some guidance for the needed translation.

Also there is a limitation on array merge that is fix only on PHP7.4 with the spread operator.


Mikescops:
It is sightly different approach than #392, same limitations indeed.
Hope we can bring back the topic on the table.

lib/AppInfo/Application.php
```diff
@@ -50,6 +53,15 @@ class Application extends App implements IBootstrap {
 		// 'video/x-matroska'	// too rarely used for photos
 	];
 
+	/** Starting PHP7.4 we will be able to use spread operator (...) to merge the 2 arrays above */
+	public const MIMES = [
+		'image/png',
+		'image/jpeg',
+		'image/heic',
+		'video/mp4',
+		'video/quicktime',
```
skjnldsv, line 62:
```suggestion
		...IMAGE_MIMES,
		...VIDEO_MIMES,
```

skjnldsv: I thought it was js xD

src/router/index.js
```diff
@@ -77,6 +77,15 @@ export default new Router({
 				showShared: true,
 			}),
 		},
+		{
+			path: '/videos',
+			component: Timeline,
+			name: 'videos',
+			props: route => ({
+				rootTitle: t('photos', 'Your videos'),
+				onlyVideos: true,
```
skjnldsv, line 86: we could also directly pass the mimes as props from here?

skjnldsv: Something like this ?
```suggestion
				mimesType: videoMimes,
```

lib/AppInfo/Application.php
Mikescops: nope but as I said PHP implemented it in 7.4, that's awesome 👍

src/router/index.js
Mikescops: good idea

skjnldsv: That way you can even do the spread from here:
```suggestion
				mimesType: [...videoMimes, ...imageMimes],
```

Mikescops: gotcha

Mikescops:
Done @skjnldsv

skjnldsv (approved):
Code looks :ok_hand: :rocket:

Mikescops:
Rebased with latest master.

jancborchardt:
It’s still a bit strange to have the videos filtered if they show no previews at all. :\ Hope we still find a solution there, also ref the discussion from @jakobroehrl’s pull request in https://github.com/nextcloud/photos/pull/392

Mikescops:
So I suggested an idea to @skjnldsv, when there is no preview, we display the name of the video in the tile. Also we can suggest with a banner a way to activate video previews

jancborchardt:
> when there is no preview, we display the name of the video in the tile. Also we can suggest with a banner a way to activate video previews

Both nice enhancements! :) The banner should only show to admins though who can actually make the change.

skjnldsv:
@Mikescops can you open issues for those? :)

Mikescops:
#508

PR #478, opened 2020-10-13
[Regression Fix] Readd timeline caching

Mikescops:
Fixing a regression following #468 after update of the vue-virtual-grid package.

The caching of the timeline is now used again. It helped simplifying some part of the logic too.


src/components/Loader.vue
```diff
@@ -0,0 +1,42 @@
...
+ -
+ - You should have received a copy of the GNU Affero General Public License
+ - along with this program. If not, see <http://www.gnu.org/licenses/>.
+ -
+ -->
+
+<template>
+	<div class="loader" icon="loading">
+		<slot name="icon">
+			<span :class="['loading']" />
```
skjnldsv, line 26:
```suggestion
			<span class="icon-loading" />
```

src/components/Loader.vue
```diff
@@ -0,0 +1,42 @@
...
+ - MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ - GNU Affero General Public License for more details.
+ -
+ - You should have received a copy of the GNU Affero General Public License
+ - along with this program. If not, see <http://www.gnu.org/licenses/>.
+ -
+ -->
+
+<template>
+	<div class="loader" icon="loading">
```
skjnldsv, line 24: `icon="loading"` this doesn't do anything, right?

Mikescops: oups, that's a test leftover

Mikescops:
@skjnldsv done

src/views/Timeline.vue
```diff
@@ -144,65 +222,25 @@ export default {
 			try {
 				// Load next batch of images
 				const files = await request(this.onlyFavorites, {
-					page: params.offset, // offset is incremented +1 by the virtualgrid lib
+					page: this.page, // offset is incremented +1 by the virtualgrid lib
```
skjnldsv, line 225:
```suggestion
					page: this.page,
```

skjnldsv (approved):
Tiny change, all good code-wise :)
One question though, why use Promise.resolve ?
Returning should be enough in those cases,

Mikescops:
The lib expect a Promise so that it can do other actions in the meantime.

skjnldsv:
> The lib expect a Promise so that it can do other actions in the meantime.

The async function resolves themselves with return or throw.
Async functions are already promises :)

skjnldsv:
This returns a promise:

```
function resolveAfter2Seconds() {
  return new Promise(resolve => {
    setTimeout(() => {
      resolve('resolved');
    }, 2000);
  });
}
```

`Promise.resolve` have no context here. I don't think that is necessary

Mikescops:
hmmm I need to better look at it

PR #479, opened 2020-10-13
[Regression Fix] Readd loading icon

Mikescops:
Fix for regression added by #468.

With the last vue-virtual-grid update we can set a loader component until the next batch is loaded.


src/components/Loader.vue
```diff
@@ -0,0 +1,21 @@
+
```
skjnldsv, line 1: copyright header ;)

skjnldsv (approved):
Minor change request, good otherwise

src/components/Loader.vue
Mikescops: Is that really needed? I added it, but one LICENSE at the root of the project looks enough to me.

skjnldsv: Well, that's how we do things here :)

src/components/Loader.vue
```diff
@@ -0,0 +1,40 @@
+<!--
+ - @copyright Copyright (c) 2020 Nextcloud
+ -
+ - @license GNU AGPL version 3 or any later version
```
skjnldsv, line 4: Nope, you're the owner!

skjnldsv (changes requested):

src/components/Loader.vue
skjnldsv: Btw please also add yourself as the author for the other files you edited

Mikescops: :disappointed:

Mikescops: done

skjnldsv (approved):

Mikescops:
@skjnldsv it was not a merge on master...

Need to review #478 first

skjnldsv:
I thought it was on purpose so I could review 478 after :grin:

PR #480, opened 2020-10-15
Fully migrate to the new Virtual Grid

Mikescops:
Migrate the last views and remove the old grid.


src/mixins/GridConfig.js
```diff
@@ -36,7 +36,7 @@ export default {
 		getGridConfig.$on('changed', val => {
 			this.gridConfig = val
 		})
-		console.debug('Current grid config', getGridConfig.gridConfig)
+		// console.debug('Current grid config', getGridConfig.gridConfig)
```
skjnldsv, line 39: hum?

skjnldsv (approved):
Works nicely!! :rocket: 
Found one issue that doesn't seems related, missing padding on top of timeline

Capture d’écran_2020-10-15_12-04-19

src/mixins/GridConfig.js
Mikescops: this log annoy me, but you left it here for long :joy: 
i just wanted to sacralize it by commented it :joy:

skjnldsv: Nah, it's here on purpose, I tend to leave debug logs
You can hide them in the console (hidden by default btw) ;)

Mikescops:
@skjnldsv that's a bug when we forget to reset the grid i think.
Can you try to reproduce it?

skjnldsv:
> Can you try to reproduce it?

It happens on initial load for me :thinking:

Mikescops:
:fearful: what, not me...

Mikescops:
Can you force refresh the page? Maybe you got cache I don't know

skjnldsv:
> Can you force refresh the page? Maybe you got cache I don't know

Well, padding is set to  0 https://github.com/nextcloud/photos/blob/f6d61acf5ca12e170ee6c78166a0bc830a0b688c/src/views/Timeline.vue#L312

Mikescops:
You should have the first title of the month that do a padding, that's what we discussed with Jan, you likely don't have a title, so it should be an issue with the js

PR #487, opened 2020-10-15
Fix svg raw loading

skjnldsv:
Signed-off-by: John Molakvoæ (skjnldsv) <skjnldsv@protonmail.com>


Mikescops (approved):
Works on my machine :D

PR #501, opened 2020-10-19
Block viewer loop on timeline

Mikescops:
On timeline going from first picture to last was not ideal as there is infinite scroll, so it was going to the last already loaded picture. 

Also this feels not natural to loop over a timeline.


skjnldsv (approved):

jancborchardt (commented):
:+1: design-wise. However on testing the branch, I can still loop around, and the "previous" button is still shown for the most recent picture? (Even after multiple force refreshes.)

Mikescops:
@jancborchardt do you have the latest viewer version?

Mikescops:
@jancborchardt FYI https://github.com/nextcloud/viewer/pull/647

jancborchardt:
> @jancborchardt do you have the latest viewer version?

That was the issue of course, always forgetting about cross-app dependencies. :) Nice work!

Issue #83, opened 2020-01-02
Multiselect actions


Mikescops:
A user has to be able to apply a certain action (delete, add to album) by selecting a batch of photos from the timeline.

There a different ways to handle that, on of them is to display a check mark when hovering the thumbnail and a click on the icon leads the user to the select mode where he can click on thumbnails to add them to the selection and then apply an action to the files from the menu.

Here is a more visual example from Google Photos:
![image](https://user-images.githubusercontent.com/4266283/71664718-508dd380-2d5a-11ea-9b85-2cd1eb590782.png)

skjnldsv:
cc @jancborchardt

jancborchardt:
Yep, multiselect and multiselect actions should absolutely be possible.

For the first step, what you describe above works:

> display a check mark when hovering the thumbnail and a click on the icon leads the user to the select mode where he can click on thumbnails to add them to the selection and then apply an action to the files from the menu.

We then also need to have a way to enter the select mode when you work on a touchscreen. It could be in a 3-dot menu next to the "Share" icon next to the folder heading.

jakobroehrl:
@skjnldsv How hard is this to do? Maybe I could do it with a little help from you?

skjnldsv:
@jakobroehrl sure thing! I don't think it's that complicated to manage for the selection
But we're lacking a proper standard to register file actions. So we'd have no way of knowing what we're allowed to do with those files unless there is one list of actions hardcoded in the app.

It would be a nice first step anyway, but it will needs to be hardcoded and when the proper standard arrive, we'd be able to use it. Are you still up for the task? :)

kanintesova:
I notice this is implemented in the mobile phone app, atleast on iOS. But no such thing on the desktop web browser.

RikkiBC:
Having to redirect web users from Photos view to File view is a little frustrating for some users, especially those coming from Google Photos (as per the original post in here). Is there any update on this issue? 
Not trying to be rude, just wondering.

r01k:
3 years later and the luxury to select multiple photos is not yet present.

joshtrichards:
This was implemented awhile ago. I believe in #1119.

PR #502, opened 2020-10-20
WIP: Implement file selector

Mikescops:
Add file selection to the grid and open sidebar when selecting a new file.

grafik


Mikescops:
I wanted to use a Set to keep track of selected files. It's very performant for such usage. But VueJS does not support Set and Map collections as reactive.
This indeed implies issues when updating the view.

Let me know if you have ideas on how we can improve this. I'm really not into using an array here tbh.

Mikescops:
@skjnldsv done with the new changes

@jancborchardt I'd like advices on the design part

jakobroehrl:
@Mikescops 
Could this be used for a multiselect actions?
https://github.com/nextcloud/photos/issues/83

jancborchardt:
Cool! Here’s some design feedback:
- [ ] The select icons look good, but don’t we want to use the Nextcloud-native ones? We can use them in monochrome (so not blue), but still the form could/should be the same?
- [ ] It’s not possible to select items using keyboard only
- [ ] The icon-checkmark shouldn’t be a span but a button, to be semantically correct?
- [ ] The checkmark should be 44*44px clickable area as per our guidelines
- [ ] For better feedback, it also should have cursor:pointer
- [ ] Possibly a border (e.g. color-primary) around the pictures is better for highlighting than tinting it, since the tint actually makes the pictures _less_ focused/visible.
- [ ] I couldn’t see any change initially until I saw the selection bar on the bottom. The selection bar would be better at the top, just like we do it for Files and Mail too. :) 
- [ ] Of course we need some actions there, if we have a primary button it will also be more visible then.
- [ ] Style-wise the selection bar should not be a hard bar, but floating buttons/text with a gradient of white to transparency down (just like the header in Mail)

Mikescops:
@jancborchardt I will answer a few remarks first.

Yes the sidebar open when you click on files (I know it may not be wanted but I was looking for a simple way to open  the sidebar and it just felt like the preview sidebar on Windows or Linux that is opened when you click on an image)

I first displayed the select bar at the bottom because if it's at the top it is hiding the navigation bar in the folders, so I don't know how you want to handle that.

Mikescops:
@jakobroehrl yes goal is to later add some action buttons, but one thing at the time.

jancborchardt:
> Yes the sidebar open when you click on files (I know it may not be wanted but I was looking for a simple way to open the sidebar and it just felt like the preview sidebar on Windows or Linux that is opened when you click on an image)

Hm, since our focus is Photos and not "a photos view of files", clicking on an image should always open the image. We could discuss that it should also always open the sidebar, but I think that having the sidebar open without also having the viewer open is not really necessary.
(Also, in the case of the sidebar we should work on the possibility to not show the preview in the sidebar, as here it’s duplication of what is shown in the viewer.)

> I first displayed the select bar at the bottom because if it's at the top it is hiding the navigation bar in the folders, so I don't know how you want to handle that.

It’s totally fine to replace the navigation bar as well, kind of like Android does. Full focus on the task at hand, which then is selection.

skjnldsv:
Implemented

Issue #499, opened 2020-10-19
Disable photos with tags view if tags app is disabled


jakobroehrl:
If the App "Collaborative tags" is not enabled the "photos with tags" view should hide

PR #503, opened 2020-10-20
show if tags are installed

jakobroehrl:
Solves: #499


src/services/AreTagsInstalled.js
```diff
@@ -0,0 +1,26 @@
+/**
+ * @copyright Copyright (c) 2018 John Molakvoæ <skjnldsv@protonmail.com>
+ *
+ * @author John Molakvoæ <skjnldsv@protonmail.com>
```
skjnldsv, line 4: It's your code :)

skjnldsv:
You need to ship the production builds :)
Capture d’écran_2020-10-20_15-00-29

jakobroehrl:
> 
> 
> You need to ship the production builds :)
> Capture d’écran_2020-10-20_15-00-29

OK, how to do this?

skjnldsv:
`npm run build` instead of `npm run dev/watch`

skjnldsv:
Or use the compile bot like this (`/compile /js`, `/compile amend /js`)
/compile /js

Mikescops (changes requested):
PR is good except that the path `apps/photos/tags` is still accessible when the app tags is disabled. Can you redirect to `/` when it's the case?

Mikescops:
Closes #499

jakobroehrl:
> 
> 
> PR is good except that the path `apps/photos/tags` is still accessible when the app tags is disabled. Can you redirect to `/` when it's the case?

Good idea. I now have the areTagsInstalled value here: https://github.com/nextcloud/photos/blob/master/src/router/index.js
How to adapt the routes?

Mikescops:
@jakobroehrl you can do something similar: isMapsInstalled

jakobroehrl:
> 
> 
> @jakobroehrl you can do something similar: isMapsInstalled

Tried these solutions, doesn't work:
https://github.com/nextcloud/photos/blob/a706e9c0bcecc07a5d3c035c2cb1b6265739a986/src/router/index.js
https://github.com/nextcloud/photos/blob/bfb72e18a9643b9150ae74f0e8c8fc6b0802a24b/src/router/index.js

What can I do?

src/router/index.js
```diff
@@ -97,7 +103,8 @@ export default new Router({
 			}),
 		},
 		{
-			path: '/tags/:path*',
+			// path: '/tags/:path*',
+			path: tagsPath,
```
Mikescops, line 107: The router path should stay `'/tags/:path*'`

Mikescops: you can add 
```
beforeEnter() {
                        if (!areTagsInstalled) {
				window.open(generateUrl('/apps/photos'))
			},
},
```

Mikescops:
@jakobroehrl I added a comment to help you

src/router/index.js
jakobroehrl: With this change the tags route isn't working any more

jakobroehrl: @Mikescops
any ideas what I can do? beforeEnter code is unfortunately not working.

Mikescops: let me checkout you're code locally

Mikescops:
@jakobroehrl here you are

jakobroehrl:
> 
> 
> @jakobroehrl here you are

Great, thank you. Now let's merge it in.

Mikescops (approved):

PR #507, opened 2020-10-21
Add recursive folder previews

### LINKED ISSUE #489: Show previews from subfolders if folder doesn't contains any images
When using the folder view, the folders that doesn't contain images are showed blank with the folder icon. They should show the folder's subfolder's images if possible

<details>
<summary>Old description (Click to expand)</summary>
Can't remember which version the change happened, but I really miss the old per-folder gallery method, instead of everything strewn together in one big list of pictures. It's really disorganised and painful to use now. I really want to see the old structure style as an option (or by default), not just how it is now. It was really annoying the change was done without considering how well the previous method worked.

Can we please get this as an option?
</details>

Mikescops:
I'm not super fan of recursivity because it can lead to lot of computation, I suggest to lock it to **one** sub level only. I guess this would fit most of the people configuration, what do you think @skjnldsv ?

Closes #489


jancborchardt:
One level seems good. Does that then also mean if you go down one level, the previews of the next sublevel will automatically start being generated already? (Which would be nice, as anticipation.)

Mikescops:
I will discuss with @skjnldsv about it.
But we don't start pre-generating inner folders

Mikescops:
@skjnldsv let me know if you have a better idea on how to do it

skjnldsv:
Do as we do for folder? We fetch the first 4 images but one level deeper?

Mikescops:
I think it's good

skjnldsv (approved):
Yveddt9

skjnldsv:
Please rebase and compile :)

BloodyIron:
I would prefer that recursivity matched the same behaviour that folder previews did before the major gallery overhaul several major versions ago. I actually used that preview recursivity to help me rapidly traverse many complex folder structures. So if this matches that previous style, I'm game for it. If not, I'd like to see that reivisited/implemented please.

Mikescops:
Not sure what you are talking about. Please provide examples, and I recommend trying current version before.

BloodyIron:
I've tried the current and previous versions, I've reported this in another issue# and was directed here. I don't know how my explanation isn't clear...

PR #514, opened 2020-10-23
Fix %2F bug in folder paths

### LINKED ISSUE #327: Albums URL bug in subfolders, %2F instead of /
When you are using the 'albums' feature (index.php/apps/photos/albums/) and clicking through folder hierarchies, sometimes the URL changes wrongly when using Firefox (75.0 on Linux Mint).
When you go to a subfolder of any folder, the URL changes to e.g. "index.php/apps/photos/albums/MyPictures%2FGreece" instead of "index.php/apps/photos/albums/MyPictures/Greece", the / is wrongly displayed as the html-code %2F. When you reload while the browser in is such a subfolder, apache just passes you a 404-error. 
When going to even deeper subfolders, the error continues, / is still displayed as %2F which results in 404-errors when reloading. 

Screenshot from 2020-05-04 20-43-14

I am not skilled enough in php and in the NextCloud backend to pass a suitable correction for the app, so it would be nice, if anybody who feels skilled enough, could fix this little bug.

Greetings, John

Mikescops:
Closes #327


src/router/index.js
```diff
@@ -58,8 +58,8 @@ export default new Router({
 			component: Albums,
 			name: 'albums',
 			props: route => ({
-				// always lead current path with a slash
-				path: `/${route.params.path ? route.params.path : ''}`,
+				// join arguments if array and always lead current path with a slash
+				path: `/${Array.isArray(route.params.path) ? route.params.path.join('/') : route.params.path || ''}`,
```
skjnldsv, line 62: Could this be in a dedicated function maybe? For a bit more clarity?

Mikescops: Sure

skjnldsv (approved):

PR #515, opened 2020-10-23
Select root folder for timeline

### LINKED ISSUE #141: Limit Photos to a single folder + subtree
I think it would be really nice to create an option where the user can select a single folder (like `/Photos`) and the app only scans this folder and its subfolders.

I  have tons and tons of images all over my nextcloud, but only in `/Photos` are my, well, photos. And it's kind of a tedious task to create always a `.nomedia` file in all other directories.

Mikescops:
Closes #141

### DIFF
--- a/lib/Controller/PageController.php
+++ b/lib/Controller/PageController.php
@@ -86,7 +86,10 @@ class PageController extends Controller {
 		$this->initialStateService->provideInitialState($this->appName, 'image-mimes', Application::IMAGE_MIMES);
 		$this->initialStateService->provideInitialState($this->appName, 'video-mimes', Application::VIDEO_MIMES);
 		$this->initialStateService->provideInitialState($this->appName, 'maps', $this->appManager->isEnabledForUser('maps') === true);
+
+		// User Settings
 		$this->initialStateService->provideInitialState($this->appName, 'croppedLayout', $this->config->getUserValue($user->getUid(), Application::APP_ID, 'croppedLayout', 'false'));
+		$this->initialStateService->provideInitialState($this->appName, 'timelineRootFolder', $this->config->getUserValue($user->getUid(), Application::APP_ID, 'timelineRootFolder', '/'));
 
 		Util::addScript(Application::APP_ID, 'photos-main');
 		Util::addStyle(Application::APP_ID, 'icons');
--- a/src/services/PhotoSearch.js
+++ b/src/services/PhotoSearch.js
@@ -44,9 +44,11 @@ export default async function(onlyFavorites = false, options = {}) {
 		page: 0, // start at the first page
 		perPage: sizes.max.count * 10, // ten rows of the max width
 		mimesType: allMimes, // all mimes types
+		rootFolder: '',
 	}, options)
 
-	const prefixPath = `/files/${getCurrentUser().uid}`
+	const davPath = `/files/${getCurrentUser().uid}`
+	const prefixPath = `${davPath}${options.rootFolder}`
 
 	// generating the search or condition
 	// based on the allowed mimetypes
@@ -65,6 +67,15 @@ export default async function(onlyFavorites = false, options = {}) {
 					<oc:favorite/>
 				</d:prop>
 				<d:literal>1</d:literal>
+			</d:eq>`
+		: ''
+
+	const eqOwner = options.rootFolder === ''
+		? `<d:eq>
+				<d:prop>
+					<oc:owner-id/>
+				</d:prop>
+				<d:literal>${getCurrentUser().uid}</d:literal>
 			</d:eq>`
 		: ''
 
@@ -97,12 +108,7 @@ export default async function(onlyFavorites = false, options = {}) {
 								${orMime}
 							</d:or>
 							${eqFavorites}
-							<d:eq>
-								<d:prop>
-									<oc:owner-id/>
-								</d:prop>
-								<d:literal>${getCurrentUser().uid}</d:literal>
-							</d:eq>
+							${eqOwner}
 						</d:and>
 					</d:where>
 					<d:orderby>
@@ -126,6 +132,6 @@ export default async function(onlyFavorites = false, options = {}) {
 	return response.data
 		.map(data => genFileInfo(data))
 		// remove prefix path from full file path
-		.map(data => Object.assign({}, data, { filename: data.filename.replace(prefixPath, '') }))
+		.map(data => Object.assign({}, data, { filename: data.filename.replace(davPath, '') }))
 
 }
--- a/package-lock.json
+++ b/package-lock.json
@@ -2554,14 +2554,24 @@
 			}
 		},
 		"@nextcloud/dialogs": {
-			"version": "2.0.1",
-			"resolved": "https://registry.npmjs.org/@nextcloud/dialogs/-/dialogs-2.0.1.tgz",
-			"integrity": "sha512-Bme8vcs8n4XT5spBgkDEv1z9zNOE23AIbr5jF1WJ1A2XNMNj5Zvy29RosIh0k7H+1lN0PlU38u+eMV1Ets3E4A==",
+			"version": "3.0.0",
+			"resolved": "https://registry.npmjs.org/@nextcloud/dialogs/-/dialogs-3.0.0.tgz",
+			"integrity": "sha512-5FVP0RSxIpKTKdSUlQ4osDDz/oCx2/4+InliB5MX2EcrjDe6q3fZMabSGnFTnIAu0CXRTzBk7RpneaIFGv+d5A==",
 			"requires": {
 				"@nextcloud/l10n": "^1.3.0",
-				"@nextcloud/typings": "^0.2.2",
+				"@nextcloud/typings": "^1.0.0",
 				"core-js": "^3.6.4",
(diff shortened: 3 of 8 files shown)


Mikescops:
@skjnldsv I don't know why but .jpg (which are JPEG files) are not taken in account, for some reasons, and as of now i can't find why..

Mikescops:
Seems like the dav:or is not working and only the first mime type is taken in account.

strugee:
This only provides for one root folder, right? That would be a problem for me and I assume other folks too since there's a couple places photos go if you use the out-of-the-box defaults - brand new accounts have a "Photos" folder created for them, so that's a natural place to put them, but the Android app auto uploads photos to the "InstantUpload" folder by default.

ntimo:
But when there is no photos directory select manually the app should function as it does with the default settings, I would assume. So this would also some your issue @strugee. Am I correct @Mikescops? :)

strugee:
@ntimo right, that would be a workaround. But then I have the original issue described in #141 again.

Mikescops:
By default no difference with the current version, but you can select one folder from where images on your timeline will be fetched from.

src/services/PhotoSearch.js
```diff
@@ -68,6 +70,15 @@ export default async function(onlyFavorites = false, options = {}) {
...
 		: ''
 
+	const eqOwner = options.rootFolder === ''
+		? `<d:eq>
+				<d:prop>
+					<oc:owner-id/>
+				</d:prop>
+				<d:literal>${getCurrentUser().uid}</d:literal>
+			</d:eq>`
+		: ''
```
skjnldsv, line 80: This is needed to not have the shares as results.
Why did you need to put a condition on it?
It should not cause any issues like it was before :)

lib/Controller/PageController.php
```diff
@@ -86,7 +86,10 @@ public function index(): TemplateResponse {
 		$this->initialStateService->provideInitialState($this->appName, 'image-mimes', Application::IMAGE_MIMES);
 		$this->initialStateService->provideInitialState($this->appName, 'video-mimes', Application::VIDEO_MIMES);
 		$this->initialStateService->provideInitialState($this->appName, 'maps', $this->appManager->isEnabledForUser('maps') === true);
+
+		// User Settings
 		$this->initialStateService->provideInitialState($this->appName, 'croppedLayout', $this->config->getUserValue($user->getUid(), Application::APP_ID, 'croppedLayout', 'false'));
+		$this->initialStateService->provideInitialState($this->appName, 'timelineRootFolder', $this->config->getUserValue($user->getUid(), Application::APP_ID, 'timelineRootFolder', '/'));
```
skjnldsv, line 92: We should probably limit the allowed keys here: https://github.com/nextcloud/photos/blob/441f97020bee10aeb9e0b27cd2ac174ed02dc3b1/lib/Controller/ApiController.php#L66

As a safety mesure :thinking:

src/services/PhotoSearch.js
Mikescops: The owner filter is only allowed on the root. Dav throws this error if you try on subfolders. What do you think then?

skjnldsv: Then it should be fixed on server.
cc @rullzer and @icewind1991 :)

skjnldsv: https://github.com/nextcloud/server/pull/17941

skjnldsv: https://github.com/nextcloud/server/blob/2a054e6c04e0a40421510eb889cbf59f153c5177/lib/private/Files/Node/Folder.php#L246

lib/Controller/PageController.php
Mikescops: I'm not sure what you mean

skjnldsv: setUserConfig allows anything, and will put it straight into the db.
Meaning we could override existing values. We should check if the $key is either `timelineRootFolder` or `croppedLayout` and throw otherwise as a safety measure

Mikescops: ok gotcha

strugee:
> By default no difference with the current version, but you can select one folder from where images on your timeline will be fetched from.

@Mikescops right, to clarify, I'm not suggesting this would break existing workflows - I'm just saying that the feature might not actually be complete enough for many(?) users due to defaults in the server/Android app. Speaking only for myself, for example, I would not be able to use this feature since given the way my Nextcloud is organized, if I turned it on I would end up missing a ton of images that I'd want to see in the Photos app.

@jancborchardt I'd be interested in hearing if you agree or if you think it's not worth the additional interface/code complexity.

Mikescops:
It's outside of the scope of this PR, if you're not happy with it, open a separate issue after it's merged, thanks.

saywebsolutions:
Which release will this go into? I'm looking forward to being able to use photos again, thank you @Mikescops

Mikescops:
NC 21 at least, for now we're stuck with a server issue.

StarSmasher44:
Hi! Sorry to bring this back, but I am hopeful this merge will resolve my issue. Is there any status update on this?
 I saw the 'server problem' pull had been merged, is there anything else that needs to be done or fixed in order to get this in asap?

skjnldsv:
> I saw the 'server problem' pull had been merged, is there anything else that needs to be done or fixed in order to get this in asap?

Hey! The server hasn't received anything regarding this fix yet. Unfortunately this is not yet possible :/

StarSmasher44:
> > I saw the 'server problem' pull had been merged, is there anything else that needs to be done or fixed in order to get this in asap?
> 
> Hey! The server hasn't received anything regarding this fix yet. Unfortunately this is not yet possible :/

Ah well that sucks, I was referring to this one: https://github.com/nextcloud/server/pull/17941 mentioned some posts up.

But would you happen to know what still needs to be done/added/fixed to make this work? I have a limited background in web development languages so I might be able to pull something through, even though I highly doubt it.

skjnldsv:
> But would you happen to know what still needs to be done/added/fixed to make this work? I have a limited background in web development languages so I might be able to pull something through, even though I highly doubt it.

It require a fair amount of rework on the backend dav search implementation :/

StarSmasher44:
> > But would you happen to know what still needs to be done/added/fixed to make this work? I have a limited background in web development languages so I might be able to pull something through, even though I highly doubt it.
> 
> It require a fair amount of rework on the backend dav search implementation :/

Sorry for being overly asking questions or coming off as pushy in any regard, I am merely just trying to figure out a good entry point, I looked through the code twice but I can't find an obvious stopping point beside the mentioned error with single vs multiple mime types being used.

And I am unsure if you even know enough of this PR yourself to respond (It isn't easy going off using other people's work as a starting point for yourself, imo) but I still have no idea what is actually wrong with this code, what needs fixing and what doesn't.

To me it looks mostly fine, given all it really needs to do is enter/search a single folder and mark that as the 'Photos' folder for NextCloud, if I get this correct.

If I have more information I may try to play around with this to see if I can bring it to a working state somehow, I have a bad taste for using OneDrive as my image backup server ever since Samsung forced it upon me, and this still adds in 600 images from my several Unity projects.

skjnldsv:
@StarSmasher44 I suggest you read the original issues like https://github.com/nextcloud/server/pull/17941 https://github.com/nextcloud/photos/issues/141

Again, this requires a fix on server. Currently there is a hard limitation for our webdav SEARCH that forces you to limit the search to a user home for performance issues.
That is why we don't have external storage nor groupfolders. And if you ignore this, you will not be able to properly filter it anymore. This requires a big revamp for al the following points to be implemented:
- Allow to search/filter by tags (favourites)
- Allow to search/filter by mime
- Allow to search/filter by user
- Allow to search/filter within a folder OR from root (currently only root is allowed with filtering)
- Allow to limit results
- Allow to sort results

Currently, the combo `search-from-folder` AND `filter-by-owner` is missing. It's not enabled (hard-coded on server) Once it is here, we'll be able to tell the Photos app to search from within the provided root folder instead of the default root.

timsayshey:
A rewrite sounds like a long term goal. Should we just stick with the original Gallery plugin until then? I think most people's Photo apps are a mix-mash of random images from all over their cloud.

> @StarSmasher44 I suggest you read the original issues like [nextcloud/server#17941](https://github.com/nextcloud/server/pull/17941) #141 entirely (with the folded comments too)
> 
> Again, this requires a fix on server. Currently there is a hard limitation for our webdav SEARCH that forces you to limit the search to a user home for performance issues.
> That is why we don't have external storage nor groupfolders. And if you ignore this, you will not be able to properly filter it anymore. This requires a big revamp for al the following points to be implemented:
> 
> * Allow to search/filter by tags (favourites)
> * Allow to search/filter by mime
> * Allow to search/filter by user
> * Allow to search/filter within a folder OR from root (currently only root is allowed with filtering)
> * Allow to limit results
> * Allow to sort results
> 
> Currently, the combo `search-from-folder` AND `filter-by-owner` is missing. It's not enabled (hard-coded on server) Once it is here, we'll be able to tell the Photos app to search from within the provided root folder instead of the default root.

AndyScherzinger:
@artonge I guess this PR is obsolete with #2319 having been merged?

PR #519, opened 2020-10-25
Add option to display specific directory only

nhirokinet:
I tried to address #141 .

In this pull request, user can set one directory to use in photo app, and photos app shows within the directory by default.

Search method is not very efficient, but it at least works in a simple way. I tried to limit scope in WebDAV search, but failed with an error.

I was not sure how you build `js/photos-*` for pushing to this directory, I kept them as is for now. For testing I used `make build-js`.

I'm not sure about coding styles here, so if you find something not fitting this repository, please comment.

### DIFF
--- a/appinfo/info.xml
+++ b/appinfo/info.xml
@@ -25,4 +25,8 @@
             <order>1</order>
         </navigation>
     </navigations>
+    <settings>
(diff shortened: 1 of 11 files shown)


nhirokinet:
Seems duplicated now. If some codes are not I'll reopen or recreate then.

PR #649, opened 2021-02-02
Update FolderTagPreview.vue

DeJQit:
This implementation provides the following UI improvement:

* Multi-line text beside folder icon

photos-css-change

### DIFF
--- a/l10n/oc.js
+++ b/l10n/oc.js
@@ -2,9 +2,24 @@ OC.L10N.register(
     "photos",
     {
     "Photos" : "Photos",
+    "Your memories under your control" : "Vòstres sovenirs jos vòstre contraròtle",
+    "Your photos" : "Vòstras fòtos",
+    "Your videos" : "Vòstras vidèos",
     "Favorites" : "Favorits",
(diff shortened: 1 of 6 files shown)


skjnldsv:
Hey!
Sorry, that was on purpose for a more pleasant look.

cc @jancborchardt for design validation

victormgarciao (approved):

Mikescops:
@jancborchardt can you review this design-wise?

marcelklehr:
I think this should be limited to two lines. That would yield the maximum of design and legibility, IMO.

jancborchardt:
Agree with @marcelklehr on limiting it to 2 lines. And line-height needs to be higher here, ideally it is somewhere between 130%–150% (best use percentage values).

skjnldsv:
> I think this should be limited to two lines. That would yield the maximum of design and legibility, IMO.

Line clamp is widely supported now, but stilla draft
https://caniuse.com/css-line-clamp

skjnldsv:
Updated in 25

PR #844, opened 2021-08-06
allow image/x-dcraw (eg .cr2) in photos

### LINKED ISSUE #530: RAW files (.CR2) not shown in Photos app but are shown in Files app.
Photos does not show previews of mimetypes from Camera RAW previews. The 'Files' app shows the previews properly (NOTE: I generated them using preview-generator), when I click on .CR2 etc.. from the 'Files' app it opens them in the preview viewer. When I switch to 'Photos' app, none of them show up (and folders which have purely CR2 are not shown at all).

 - OS: Raspberry pi stretch
 - Browser Chrome
 - Version 86

boecko:
This fixes #530.
.cr2 files are now availabe in photos.

### DIFF
--- a/lib/AppInfo/Application.php
+++ b/lib/AppInfo/Application.php
@@ -37,6 +37,7 @@ class Application extends App implements IBootstrap {
 		'image/png',
 		'image/jpeg',
 		'image/heic',
+		'image/x-dcraw'
 		// 'image/gif',			// too rarely used for photos
 		// 'image/x-xbitmap',	// too rarely used for photos
 		// 'image/bmp',			// too rarely used for photos


skjnldsv (approved):

skjnldsv:
Aaaah, I was too fast.
I forgot that those previews are actually provided by the https://github.com/ariselseng/camerarawpreviews app.
Meaninng if it's not installed, the photos app will wtill display the CR2 files, so that's not good.
What we need is https://github.com/nextcloud/server/issues/22873

So we can get the list of mimes that nextcloud supports, instead of relying on this hardcoded list.

PR #894, opened 2021-09-30
New Gallery view for photos app

TSI-ashishsharma:
Signed-off-by: Ashish Sharma <sharmaa@t-systems.com>

Adding a new gallery view with one checkbox condition.
if the checkbox is checked then it will automatically load the gallery view instead of the normal photos app view.


AndyScherzinger:
Hi @TSI-ashishsharma, thanks for the PR, could you please add some screenshots, this make it easier for the designers to check the UI/UX changes.
Thanks in advance ❤️

TSI-ashishsharma:
> Hi @TSI-ashishsharma, thanks for the PR, could you please add some screenshots, this make it easier for the designers to check the UI/UX changes. Thanks in advance heart

Screenshots added in the screenshot folder.  please verify :)

skjnldsv:
Quick overall comment.
We're using a wrapper to display every photos called `VirtualGrid`
https://github.com/nextcloud/photos/blob/88351ebc3c8caa5ca388552a600ecf92caeb5e7f/src/views/Timeline.vue#L47-L54
https://github.com/nextcloud/photos/blob/88351ebc3c8caa5ca388552a600ecf92caeb5e7f/src/views/Timeline.vue#L68

Adding a new view should only change that component. You will need to implement or use an external npm library dedicated to fancy flex-grid.
And then you can use Vue `v-if/v-else` to just change what component is being used (VirtualGrid` or your new `FlexGrid` component or library)

But you should still keep the same behaviour or something on how pagination and/or Files display is being handled.

Maybe it's not even needed, maybe you can just implement that with Flex and calculate how many entries are being displayed or not for pagination. Your worst enemy here is virtualisation of elements. Because scrolling 10 times will give you a crazy amount of photos, if you don't remove the old from the dom, you will freeze any browser trying to render such amount of photos. You will need to find a solution that manage your display as well as virtual scrolling. Whetever s being displayed inside should then not change (aka, keep using the https://github.com/nextcloud/photos/blob/master/src/components/File.vue component)

skjnldsv:
> Screenshots added in the screenshot folder. please verify :)

Also, screenshots should be added to the issue, not in the code base ;)

PR #937, opened 2021-12-16
Implement 'On this day'

### LINKED ISSUE #179: Show pictures taken previous years on this day
I personally find it entertaining to know what i was doing on the current day in past years, ie. Browsing through my images from 6.feb 2018, 2017, 2016 and so on. I'm aware that snapchat has a feature like this with saved images, but it is limited to snaps that you upload to their servers.

Wouldn't it be a nice addition to Nextcloud photos for a function like this as well?

I'm assuming this would be possible by leveraging the EXIF-data if it's present, with a fallback to the modified date (or only focus on the modified date)

marcelklehr:
fixes #179

image

- Uses the standard Timeline component
- Goes back 20 years
- Displays photos from this day and "tomorrow" for convenience (so overnight events still work as expected)


jakobroehrl:
Yeah, freaking cool! Waited so long for this!
What do you think of 
1. extend the range to 1-2 days in the past and 1-2 days in the feature from now?
2. if there is no catch extend the range even more. Show some photos from the month of the past years e.g.? https://github.com/photoprism/photoprism/issues/720#issuecomment-762403992
3. call it moments or recommendations?

jakobroehrl:
What do you think about a notification (off per default) if there was a photo found of this day in the past?

marcelklehr:
> Waited so long for this!

Merry Christmas :christmas_tree: :santa: 

> 1. extend the range to 1-2 days in the past and 1-2 days in the feature from now?
> 2. if there is no catch extend the range even more. Show some photos from the month of the past years e.g.? https://github.com/photoprism/photoprism/issues/720#issuecomment-762403992

We can definitely discuss the date range. I'm not particular about this.

> 3. call it moments or recommendations?

Not particular about the name either. :)

// cc @nextcloud/designers

jakobroehrl:
> > What do you think about a notification (off per default) if there was a photo found of this day in the past?
> 
> I'm more in favor of a dashboard widget, personally :)

Good idea, it's better with a dashboard widget

jakobroehrl:
Would it make sense to give access via API?
This app could show the contents eg: https://gitlab.com/nkming2/nc-photos/-/issues/51
It already shows faces if enabled on the server

marcelklehr:
This PR doesn't do anything too complicated that is not already available via WebDAV. In fact, it uses the normal webdav search.

src/services/PhotoSearch.js
```diff
@@ -67,6 +70,23 @@ export default async function(onlyFavorites = false, options = {}) {
 			</d:eq>`
 		: ''
 
+	const onThisDay = options.onThisDay
+		? `<d:or>${Array(20).fill(1).map((_, years) => `<d:and>
+				<d:gt>
+					<d:prop>
+						<d:getlastmodified />
```
Mikescops, line 77: the behavior will be weird, cause lastmodified is impacted when you tag an image for instance, so it won't be exactly "on this day"

but I guess that's the best we can do with Dav

Mikescops (approved):

src/services/PhotoSearch.js
marcelklehr: Once https://github.com/nextcloud/server/pull/30366 is merged this could be changed to creation_time

jakobroehrl:
@marcelklehr 
Do you create the dashbaord widget for pictures in this PR, too?

marcelklehr:
> Do you create the dashbaord widget for pictures in this PR, too?

Nope.

src/services/PhotoSearch.js
Mikescops: ah interesting, I always thought this would cause performance issues

@skjnldsv am I wrong?

jancborchardt:
Awesome @marcelklehr! So design-wise, I was thinking it could be in the first row of "Your photos", similarly to the recommendations in Files, or how we sort favorites up top. But after a very quick mock-up, I’m actually not sure anymore. :D 
image


I’d say it’s good how you did it – just thoughts about naming and timespan:
- The headings might be better as: "**1 year ago** December 2020", otherwise it’s always December December December etc. ;)
- The naming is "On this day", but we should probably also fuzzily include photos a few days around the date, right? Cause it’s probably rare for people to take a photo every single day, and then often "1 year ago" would just be nothing. Maybe then even we rename it to "Through the years" or "Memories" or something similar?
- I assume that it’s starting in 2021 is just to show as an example? Cause it would start at 1 year ago, right?

marcelklehr:
Thanks for the review!

> The headings might be better as: "1 year ago December 2020", otherwise it’s always December December December etc. ;)

Fixed

    The naming is "On this day", but we should probably also fuzzily include photos a few days around the date, right? Cause it’s probably rare for people to take a photo every single day, and then often "1 year ago" would just be nothing. Maybe then even we rename it to "Through the years" or "Memories" or something similar?
    I assume that it’s starting in 2021 is just to show as an example? Cause it would start at 1 year ago, right?


New screenshot:

Spartachetto:
First of all thank you very much, cool feature. 

A long term suggestion: once https://github.com/nextcloud/server/issues/30416 will be included, some other possibilities could be considered. For example using the geo data and showing the  photo farthest from your usual location

jancborchardt:
Nice @marcelklehr! :) Only one detail, the heading bolding could be improved to what I mentioned:

> **1 year ago**  ·  January 2021

Specifically the "x years ago" part bold, nd the "Month 20xx" part not bold.

Additionally to what I mentioned before, for spacing there could be a middle dot inbetween with 2 spaces to either side.

marcelklehr:
> Specifically the "x years ago" part bold, nd the "Month 20xx" part not bold.

Ah. I missed that.

Here we go:
image

jancborchardt (approved):
Looks awesome! :) Great work @marcelklehr

marcelklehr:
> This branch has conflicts that must be resolved

:thinking: 

@skjnldsv Did I do something wrong?

skjnldsv:
> > This branch has conflicts that must be resolved
> 
> thinking
> 
> @skjnldsv Did I do something wrong?

No, I had to merge another big PR bumping dependencies, you need to rebase and drop js changes.

skjnldsv:
Rebased

skjnldsv:
> Great feature :) what do you think about implementation of a plugin to dashboard showing a preview from 'on this day'? I would love it as this is my main landing page when working with Nextcloud :)

Could you open an issue about this? :)
Great idea!

count00zero:
It seems to be pushed to "master", but is there an additional step to enable it in the standard photos-app within NC 23.0.2?
Cause I can't "see" the feature ...

jakobroehrl:
It will be released with the next feature release, this is 24.
23.0.2 is only for bugfixing

Issue #908, opened 2021-10-11
Photos lists items from External storage even with usage of .nomedia and .noimage empty file


dashbash9787:
**Describe the bug**
Photos lists items from External storage even with usage of .nomedia and .noimage empty file.

2 x SMB External storage

Both root dirs have .nomedia and .noimage

However both list any media files in Photos

Have rescanned both using occ files:scan -all  --path=/USERNAME/files/FOLDERNAME(as listed on External storage page)

**To Reproduce**
1. Connect to External storage using SMB
2. Photos will list items from those dirs ignoring .nomedia and .noimage

**Expected behavior**
Would expect any media in those dirs and below to be ignored.

**Desktop (please complete the following information):**
 - Operating System: Linux 5.10.28-Unraid x86_64
 - Browser: Chrome , Vivaldi
 - Nextcloud 22.2.0

hobbsecky:
this problem is effecting ALL folders that have .nomedia set

hobbsecky:
actually removing this app from my nextcloud servers. I don't see a need for it.

marcelklehr:
support for .nomedia files has landed.

Issue #903, opened 2021-10-03
Photos still shoing pictures outside my Photos folder in the root directory.


noobscode:
The photo viewer is still showing photos that I have zipped away, photos outside my photos folder. I have executed the file scan job several times and tried adding .nomedia / .noimage filder in directory but it does nothing. same issue on both pc and phone.

hobbsecky:
me too

martabal:
Same...

iluvatyr:
same... needs to be fixed for sure!

marcelklehr:
Support for .nomedia files has landed since.

Issue #537, opened 2020-11-08
.nomedia / .noimage is not working in 20.0.0


tolotos-rgu:
The webpage gallery of NC 20.0.0 ignore the .noimage / .nomedia files completely.
IOS app shows the same issue and devs said it was no app function.
As it works in NC <= 19.x it must be NC 20 server related.

skjnldsv:
Please search for existing issues before posting
https://github.com/nextcloud/photos/issues/234

skjnldsv:
Regarding the iOS app, this have to me handled with the iOS developer and repository please :)

Issue #245, opened 2020-03-19
.noimage .nomedia has no effect


bendschs:
images in folders or subfolders with .noimage or .nomedia files are still shown in the gallery.

Nextcloud 18.0.2
Same behaviour in various browsers and in the IOS-app.

skjnldsv:
Hey :)
Please do a research before posting https://github.com/nextcloud/photos/issues/234 https://github.com/nextcloud/photos/pull/159 https://github.com/nextcloud/photos/issues/75

bendschs:
ok, so no fix yet?

PR #939, opened 2021-12-30
Add support for ignoring folders with .nomedia and .noimage

### LINKED ISSUE #203: Honor .nomedia file or similar
Currently (afaik) all images are beeing shown. That leads to some images beeing shown twice when they are in a thumbnail folder. There could be a way to automatically hide folders/images that are in a folder beginning with a dot (.thumbnails) or if the folder contains a .nomedia file. This could be toggleable, like with 'show hidden files' in the files app. 

A more elaborate way would be to manage exclusions via gui so that a user could choose a folder to be excluded, but imho honoring a nomedia file or the preceeding dot should suffice.

jonjomckay:
This uses an event listener to record the paths of folders that contain a `.nomedia` or `.noimage`, and stores them as "excluded paths" in the database.

These excluded paths are then used to filter the view appropriately. Performing the filtering in the view is quite naive, but seems to be the best way for now, due to the limitations of using the WebDAV `searchrequest` function. If the app ends up moving to an app-specific API in the view, these paths can be used easily in the server-side API method, as they are.

Creation of an "included paths" table with the same schema is also included, which could be used to implement #141.

This is my first PR against anything Nextcloud-related, so let me know if I need to change anything!

This should fix #203, #234, #245, #537, #903 and #908.


jonjomckay:
One thing I'm not sure about - should I add `uid` columns to the database tables? They _should_ already be "namespaced" due to the full path being used...

jancborchardt:
FYI regarding this pull request since @marcelklehr you also opened one regarding .nomedia filtering. cc @skjnldsv

marcelklehr:
@jonjomckay I only see event listeners for the nomedia bookkeeping in the code. How does this PR deal with already existing files?

T-bond:
I think this can be closed, as #944 got merged.

PR #944, opened 2022-01-02
Implement .nomedia/.noimage filter

### LINKED ISSUE #234: .nomedia / .noimage file not respected within "Your photos"
A partial fix was implemented in https://github.com/nextcloud/photos/pull/159 for respecting .nomedia / .noimage files in a path, but unfortunately, it only applies to the "Your albums" tree.

The "Your photos" tree is still a disaster.

marcelklehr:
for Timeline and Tags. Happy new year :sparkles: 

fixes #234 

The main difference to #939 is that that one maintains a DB table with excluded paths which is the loaded on the client using initialState, while this PR ~uses a DAV SEARCH for .nomedia/.noimage files on initial page load~ uses a cached file search loaded on the client using initialState.


src/Photos.vue
```diff
@@ -102,7 +103,7 @@ export default {
 		}
 	},
 
-	beforeMount() {
+	async beforeMount() {
```
skjnldsv, line 106: I don't think vue supports async lifecycle hooks :thinking:

skjnldsv (commented):
Hi, I'm afraid such request will have an insane load on instances with high number of files.
Do you have some rough benchmarks on how long this search takes? :)

Happy new year Marcel!! :tada: :hugs:

src/Photos.vue
marcelklehr: AFAIK, Vue doesn't care because it doesn't use the return value. An async function is just syntactic sugar for a function that returns a promise, after all.

marcelklehr:
>  I'm afraid such request will have an insane load on instances with high number of files.
Do you have some rough benchmarks on how long this search takes? :)

Mh, I see, that would be bad. But why would this specific DAV search query take longer than e.g. the "Your Photos" search query?

marcelklehr:
### Benchmark

I loaded ~10.000 files into my dev instance. Not sure if that's enough, big instances are probably considerably larger, but it's something.

```
MariaDB [nextcloud]> select COUNT(*) from oc_filecache;
+----------+
| COUNT(*) |
+----------+
|     9716 |
+----------+
1 row in set (0.027 sec)
```

The DAV search for .nomedia/.noimage files with infinite depth for the current user takes ~247ms while the normal Timeline DAV search takes ~200ms.

skjnldsv:
> The DAV search for .nomedia/.noimage files with infinite depth for the current user takes ~247ms

That is quite high. It means we'd have to wait for quite some time before we get the final list of files :thinking:

marcelklehr:
> It means we'd have to wait for quite some time before we get the final list of files 

As I understand it the two queries should be executed in parallel, because Vue doesn't wait for the created hook to finish, so the load time is increased by 50ms in my setup.

I'm happy to close this PR in favor of #939, though, as long as we can land .nomedia filtering. :)

skjnldsv:
Maybe it would be best to have a private API for that. Just returning the ignored paths?

marcelklehr:
> Maybe it would be best to have a private API for that. Just returning the ignored paths?

@skjnldsv That would go into the direction of https://github.com/nextcloud/photos/pull/939 if I understand you correctly. Although, #939 maintains a database table for ignored paths. A compromise between the two PRs would be to have an initial state that does the same query as the SEARCH request in this PR?

skjnldsv:
> A compromise between the two PRs would be to have an initial state

Either initial state or dedicated request, both are fine.
if the request is expensive for the server, I would rather have this on-demand than on photos app init.

The database extra seems a bit overpowered, maybe it's worth doing a bit of in-depth analysis on how long and expensive a php search request would be. Then we can maybe drop the table idea. :shrug:

marcelklehr:
> in-depth analysis on how long and expensive a php search request would be

I'd defer to @icewind1991 on this, but from what I understand, there's the file cache table in the database which lists all files in the instance and fetching all .nomedia files from that should be fairly fast, be it using a WebDAV SEARCH request or a custom SQL query:

```
MariaDB [nextcloud]> explain select * from oc_filecache JOIN oc_mounts ON oc_filecache.storage = oc_mounts.storage_id where name = '.nomedia' and oc_mounts.user_id = 'admin';
+------+-------------+--------------+------+-----------------------------------------------------------------------------------------------------+------------------------+---------+--------------------------------+------+-----------------------+
| id   | select_type | table        | type | possible_keys                                                                                       | key                    | key_len | ref                            | rows | Extra                 |
+------+-------------+--------------+------+-----------------------------------------------------------------------------------------------------+------------------------+---------+--------------------------------+------+-----------------------+
|    1 | SIMPLE      | oc_mounts    | ref  | mounts_user_root_index,mounts_storage_index                                                         | mounts_user_root_index | 258     | const                          | 1    | Using index condition |
|    1 | SIMPLE      | oc_filecache | ref  | fs_storage_path_hash,fs_storage_mimetype,fs_storage_mimepart,fs_storage_size,fs_storage_path_prefix | fs_storage_mimetype    | 8       | nextcloud.oc_mounts.storage_id | 1552 | Using where           |
+------+-------------+--------------+------+-----------------------------------------------------------------------------------------------------+------------------------+---------+--------------------------------+------+-----------------------+
```

skjnldsv:
> On-demand is hard to do, IMHO, because we don't know whether there are any .nomedia files until we've fetched the list, right?

you can make that part of the  async/await process.
run both in parallel and filter out paths? :thinking:

marcelklehr:
A compromise would be to cache the result of searching for .nomedia files in PHP, based on the [Etag](https://nextcloud-server.netlify.app/classes/ocp-files-fileinfo#method_getEtag) of the user's root folder. Then we'd only have the performance hit of the SQL query once, in case the file tree of the current user has changed, but we wouldn't have the overhead of adding extra db tables and a file system hook.

@skjnldsv What do you think?

marcelklehr:
:shrug:

marcelklehr:
@icewind1991 Thanks for getting back to me on this :) Do you think my current approach (cached file cache search based on user folder Etag) makes sense (especially performance wise)? How would you recommend to handle this?

skjnldsv:
Looks great now @marcelklehr :)
I don't know where we should document this though. Having `.nomedia` is super technical. The best would be to have a dedicated way to select paths to exclude or #141.
Documenting would maybe makes it a bit too visible? Any opinions?

marcelklehr:
> Documenting would maybe makes it a bit too visible? Any opinions?

We could add a note to the README, maybe, but I wouldn't add it to the app UI

szaimen:
> /rebase

I fear you need to rebase manually:

```
git pull --rebase origin master
```
and then
```
git push --force-with-lease
```

marcelklehr:
@szaimen party time :tada: :laughing:

marcelklehr:
Do you want me to fix the psalm errors?

lib/Controller/PageController.php
```diff
@@ -89,6 +120,25 @@ public function index(): TemplateResponse {
...
 		$this->initialStateService->provideInitialState($this->appName, 'systemtags', $this->appManager->isEnabledForUser('systemtags') === true);
 
+		$paths = [];
+		try {
+			$userFolder = $this->rootFolder->getUserFolder($user->getUID());
+			$key = $user->getUID() . ':' . $userFolder->getEtag();
+			$paths = $this->nomediaPathsCache->get($key);
+			if ($paths === null) {
+				$nomedia = $userFolder->search('.nomedia');
+				$noimage = $userFolder->search('.noimage');
```
icewind1991, line 130: You can search this in one go with

```
$userFolder->search(new SearchQuery(new SearchBinaryOperator(ISearchBinaryOperator::OR, [
    new SearchComparison(ISearchComparison::COMPARE_EQ, 'name', '.nomedia'),
    new SearchComparison(ISearchComparison::COMPARE_EQ, 'name', '.noimage')
]), 0, 0, [], $user));
```

that also makes it not use a wildcard needlessly.

marcelklehr: > SearchQuery SearchBinaryOperator SearchComparison

Are these in the OCP namespace, though?

icewind1991: No, interfaces for them are though, but I never got around to making a query builder for it.

marcelklehr: @skjnldsv Are you ok with that?

marcelklehr:
ping @skjnldsv

lib/Controller/PageController.php
szaimen: @skjnldsv I think this needs a simple 👍 or 👎

skjnldsv: Yes, let get this moving forward :)

src/store/files.js
```diff
@@ -88,6 +103,16 @@ const actions = {
...
 	},
+
+	/**
+	 * Set list of all .nomedia/.noimage files
+	 *
+	 * @param {object} context the store mutations
+	 * @param {Array} paths list of files
+	 */
+	setNomediaPaths(context, paths) {
+		context.commit('setNomediaPaths', paths)
```
skjnldsv, line 114: Could you debug this in the console?
```suggestion
		logger.debug('Ignored paths', { paths })
```

marcelklehr: Like this?

```js
    setNomediaPaths(context, paths) {
		console.debug('Ignored paths', { paths })
		context.commit('setNomediaPaths', paths)
	},
```

skjnldsv (approved):
Looks good! :+1:

ntrp:
How I can see in which release is this coming? I was waiting for something like this since 2 years *_*

ParaplegicRacehorse:
Are these (.noimage/.nomedia) supposed to be files resident in the filesystem, or folder tags?

marcelklehr:
Simply create an empty file in the file system

ronny332:
does not work for me. has the files system wie crawled again after putting the file in the folder? I can't find the point where the `scanFolder()` method from `AlbumsController.php` runs to find out if the file exists or not. It seems not to be started at any time.

PR #946, opened 2022-01-03
Fix Tags: Don't display tags without photos

### LINKED ISSUE #747: Faeture Tags
Hello,

Congratulations on this great job. I want to know if it is possible to manage Tags by user. I found the shared tags. Can I give the possibility to users to have their own tags?

Thank you

marcelklehr:
Currently, people will see tags that they don't use and haven't set on any of their files, which is confusing. This PR removes tags without photos from the view. As you can see in the screencast, this is kind of ugly, since the systemtags endpoint doesn't tell us how many items there are for each tag (AFAIK) before we're rendering the tags.

Peek 2022-01-03 16-09

fixes #747 (YMMV)


skjnldsv:
Rebased

PR #1045, opened 2022-02-28
Unify preview urls with the files app

CarlSchwan:
This has two advantages:

* Increase the probability of browser cache hits
* With the imaginary pipeline PR, requesting a preview of 250x250 is faster
  since there are less intermediate steps. The resulting images send by the
  server is still the same one.

Signed-off-by: Carl Schwan <carl@carlschwan.eu>


src/components/File.vue
```diff
@@ -94,7 +94,7 @@ export default {
 			return this.item.injected.mime.startsWith('image')
 		},
 		src() {
-			return generateUrl(`/core/preview?fileId=${this.item.injected.fileid}&x=${256}&y=${256}&a=${!this.croppedLayout}&v=${this.item.injected.etag}`)
+			return generateUrl(`/core/preview?fileId=${this.item.injected.fileid}&c=${this.item.injected.etag.replace('&quot;', '').replace('&quot;', '')}&x=${250}&y=${250}&forceIcon=0&a=${this.croppedLayout ? '0' : '1'}`)
```
skjnldsv, line 97: This should be a method or something :grin: 

`this.item.injected.etag.replace('&quot;', '').replace('&quot;', '')`

PR #1112, opened 2022-06-10
Timeline: Do not take VirtualGrid out of DOM to avoid infinite loop

### LINKED ISSUE #1081: Photos app in NC 24.0.0 RC1 using version Photos 1.6.1 keeps loading more thumbnails
**Describe the bug**
When viewing the Photos tab, this normally loads the most recently modified images on the site. As you scroll down the page, more images are loaded up.

In the latest version of 24.0.0 RC1 I noticed that the page performance was extremely sluggish in comparisson to the previous 23 versions. 

After monitoring the page, it turns out this is because it's stuck in an infinite loop of requesting more images, making the page performance eventually poor.

Checking the Network tab in Developer Tools gives me a constant series of requests to the dav/ enpoint requesting the SEARCH metadata URL. 

**To Reproduce**
Steps to reproduce the behaviour:
1. Go to the Photos menu(?) along the top of the NC page.
2. Leave the page be and watch the scrollbar on the side of the page. This continues to grow as more images are requested.
3. Load up F12 developer tools and monitor the Network tab, the requests to the dav/ API occur about once every second.

**Expected behaviour**
I would expect this to request further images only when I scroll down, rather than constantly.

**Screenshots**

 - OS: Windows 10
 - Browser: Chrome v100, also tried in Firefox.
 - Version: 24.0.0 RC1

marcelklehr:
fixes #1081


Mikescops (approved):
:tada:

skjnldsv (approved):

skjnldsv:
I guess it's only 24

marcelklehr:
Yep, I think so

PR #1119, opened 2022-06-15
Introduce tiled layout in the timeline

artonge:
:heavy_check_mark:  **Done and ready for reviews** :heavy_check_mark: 

## Done

- [x] Finish AppNavigation rework
- [x] First draft of tiled algo
- [x] Thumbnails for videos
- [x] Allow selection
  - [x] Improve `:hover` style for Checkbox without label
  - [x] Allow setting a background to Checkbox
  - [x] Checkbox rendering on slim photos
  - [x]  Download
  - [x]  Favorite
  - [x]  Delete (3 actions in the 3 dot menu)
  - [x] Added focused state
  - [x] Select on keyuo.enter when focused
- [x] Improve reload on bottom reached
- [x] Split into section by month
- [x] Check that it is pixel perfect
  - [x] Remove slight border radius on picture
  - [x] Check that the default height is fine
- [x] Add Header
  - [x] Fix respect of `force-title` attribute in `Actions`
- [x] Clean `TiledLayout` code
  - [x] Extract algo into JS
  - [x] Write tests
  - [x] Check with very tall pictures
  - [x] Check with very large pictures
  - [x] Support resize
    - [x] Detect navigation bar toggle
  - [x] Do not flash loading indicator when fast scrolling
- [x] Implement virtual scrolling
  - [x] Support different container (window, custom element, and inner element)
- [x] What about files that do not have metadata? => default to 256*256
- [x] Next and previous in Viewer don't match image order
- [x] Prioritize loading of visible pictures
- [x] Accessibility work
- [x] Check scroll on mobile
- [x] Use `ul/li` for item rendering

## Potential follow-up tasks

- [ ] Allow (un)select all (also by section) ?
    - [ ] Select new loaded items if section is selected
    - [ ] Not loaded items should be checked on favorite, delete, ...
- [ ] Fasten viewer opening
  - [ ] Darken screen faster
  - [ ] Or use preview file in viewer instead of showing the loader
- [ ] Allow Click and slide selection (like in a classic file viewer)
- [ ] Favorite state ?
- [ ] Improve navigation in time (allow navigating to precise date instead of having to scroll)
- [ ] Static section header on scroll
- [ ] Fancy scroll bar with date for quick navigation

# Example (28/07 - 10:30)

Screenshot 2022-07-28 at 10-31-59 Photos - Nextcloud


jancborchardt:
Nice! :)
The only weird part is in January 2021 with the 2 huge pictures – but I understand why it is like that with the panorama following them.

Also, the month/date titles can use a bit more padding to the top of them. :)

src/services/TiledLayout.js
```diff
@@ -0,0 +1,202 @@
...
+ * @property {number} width Real width of the item.
+ * @property {number} height Real height of the item.
+ * @property {Vue.Component | string} renderComponent The Vue component used to render the item.
+ */
+
+/**
+ * @typedef {object} TiledRow
+ * @property {TiledItem[]} items -
+ * @property {number} height -
+ */
```
skjnldsv, line 37: We could go typescript if you want :)

src/mixins/VirtualScrolling.js
```diff
@@ -0,0 +1,153 @@
...
+ *             },
+ *         }
+ *     </script>
+ *     <style>
+ *         .vs-container {
+ *             height: 500px; // Static height
+ *             overflow-y: scroll;
+ *         }
+ *     </style>
+ */
```
skjnldsv, line 58: All of this should most likely be a direct component with a render function. We would gain a lot of performances :)

artonge:
Changed target branch to `photos2` so we can do the remaining work there.

src/components/PhotosHeader.vue
```diff
@@ -0,0 +1,202 @@
...
+		},
+
+		openAlbumPicker() {
+
+		},
+
+		async moveSelectionToAlbum() {
+
+		},
+
```
Pytal, line 137:
```suggestion
```

src/components/TiledLayout.vue
```diff
@@ -0,0 +1,125 @@
...
+export default {
+	name: 'TiledLayout',
+
+	components: {
+		VirtualScrolling,
+		Loader,
+	},
+
+	props: {
+		/** @type {Array} */
```
Pytal, line 54:
```suggestion
```

src/components/PhotosHeader.vue
artonge: I'll keep those, but commented, so it can easily be reused in future work that should happen soon.

src/components/PhotosHeader.vue
```diff
@@ -0,0 +1,201 @@
...
+ -
+ - You should have received a copy of the GNU Affero General Public License
+ - along with this program. If not, see <http://www.gnu.org/licenses/>.
+ -
+ -->
+<template>
+	<div class="photos-header">
+		<Actions v-if="selection.length === 0" :menu-title="t('photos', 'Add')" :primary="true">
+			<Plus slot="icon" />
+			<!-- <ActionButton :close-after-click="true" :title="t('photos', 'Upload media')" @click="openUploader">
```
Pytal, line 26: **suggestion (non-blocking):** Note in an inline comment which parts are placeholders explicitly for other reviewers

src/components/File.vue
```diff
@@ -21,108 +21,203 @@
...
+			<div v-if="item.mime.includes('video') && item.hasPreview" class="icon-video-white" />
+
+			<!-- image and loading placeholder -->
+			<div class="images-container">
+				<img v-if="visibility !== 'none' && canLoad && !error"
+					ref="imgNear"
+					:key="`${item.basename}-near`"
+					:src="srcNear"
+					:alt="item.basename"
+					:aria-describedby="ariaUuid"
```
Pytal, line 45:
```suggestion
					:aria-describedby="ariaId"
```
**suggestion (blocking):** If `fileid` is not a UUID compliant string this should be renamed to something like `ariaId`

src/components/VirtualScrolling.vue
```diff
@@ -0,0 +1,247 @@
...
+
+			let currentRowTopDistanceFromTop = 0
+			let currentBottomTopDistanceFromTop = 0
+
+			// Compute whether a row should be included in the DOM (shouldRender)
+			// And how visible the row is.
+			return this.rows
+				.map(
+					/**
+					 * @param {Row} row -
```
Pytal, line 108:
```suggestion
					 * @param {Row} row
```

src/services/FileActions.js
```diff
@@ -0,0 +1,105 @@
...
+ * Delete a file
+ *
+ * @param {number} fileName - The file's id
+ */
+export async function deleteFile(fileName) {
+	try {
+		await client.deleteFile(`/files/${getCurrentUser()?.uid}/${fileName}`)
+	} catch (error) {
+		logger.error(t('photos', `Fail to delete ${fileName}.`), error)
+		showError(t('photos', `Fail to delete ${fileName}.`))
```
Pytal, line 42:
```suggestion
		logger.error(t('photos', 'Failed to delete {fileName}.', { fileName }), error)
		showError(t('photos', 'Failed to delete {fileName}.', { fileName }))
```

src/services/FileActions.js
```diff
@@ -0,0 +1,105 @@
...
+	while (encodedPath[0] === '/') {
+		encodedPath = encodedPath.substring(1)
+	}
+
+	try {
+		return axios.post(
+			`${generateUrl('/apps/files/api/v1/files/')}${encodedPath}`,
+			{
+				// contentType: 'application/json',
+				// dataType: 'json',
```
Pytal, line 63:
```suggestion
```

src/services/FileActions.js
```diff
@@ -0,0 +1,105 @@
...
+				tags: favoriteState ? ['_$!<Favorite>!$_'] : [],
+			},
+		)
+	} catch (error) {
+		logger.error(t('photos', `Fail to favorite ${fileName}.`), error)
+		showError(t('photos', `Fail to favorite ${fileName}.`))
```
Pytal, line 69:
```suggestion
		logger.error(t('photos', 'Failed to favorite {fileName}.', { fileName }), error)
		showError(t('photos', 'Failed to favorite {fileName}.', { fileName }))
```

src/components/File.vue
artonge: Nice catch, I renamed it to ariaDescription for even more clarity.

src/services/FileActions.js
artonge: Thanks

src/services/FileActions.js
Pytal: Not resolved?

Pytal (approved):
Other than comments and eslint errors

jancborchardt (commented):
Looks suuper nice @artonge, only 2 details:
- [x] The date divider seems to not left align with the images, but is a bit further to the left?
- [x] The date divider can use some more whitespace to the top (make it double :)

src/services/FileActions.js
artonge: Ah ! Though it was Github rendering twice the same review :see_no_evil: Fixed now.

Pytal (approved):

src/utils/semaphoreWithPriority.js
```diff
@@ -0,0 +1,79 @@
...
+ *
+ * See the COPYING-README file.
+ *
+ */
+
+/**
+ *
+ * @param {number} capacity - The number of simultaneous access to the ressource.
+ */
+export default class SemaphoreWithPriority {
```
artonge, line 15: https://www.npmjs.com/package/p-queue ?

src/store/files.js
```diff
@@ -39,9 +44,23 @@ const mutations = {
...
+
+			// Precalculate dates as it is expensive.
+			file.timestamp = moment(file.lastmod).unix() // For sorting
+			file.month = moment(file.lastmod).format('YYYYMM') // For grouping by month
+			file.day = moment(file.lastmod).format('MMDD') // For On this day
 		})
+
+		state.files = {
+			...state.files,
+			...files.reduce((files, file) => ({ ...files, [file.fileid]: file }), {}),
```
artonge, line 62: Rename files to `newFiles`

skjnldsv (approved):

src/store/files.js
```diff
@@ -20,6 +20,11 @@
  *
  */
 import Vue from 'vue'
+import moment from '@nextcloud/moment'
```
artonge, line 23: locale moment ?

skjnldsv:
LGTM ? :rocket:

src/components/File.vue
```diff
@@ -21,108 +21,203 @@
...
-			</svg>
-		</transition-group>
-
-		<!-- image name and cover -->
-		<p :id="ariaUuid" class="hidden-visually">{{ item.injected.basename }}</p>
-		<div class="cover" role="none" />
-	</a>
+	<div class="file-container"
+		:class="{selected}">
+		<CheckboxRadioSwitch v-if="allowSelection"
```
CarlSchwan, line 26: Accessibility: checkbox without visible or invisble label

src/components/TiledLayout.vue
```diff
@@ -0,0 +1,101 @@
...
+		this.resizeObserver.observe(this.$refs.tiledLayoutContainer)
+	},
+
+	beforeDestroy() {
+		this.resizeObserver.disconnect()
+	},
+}
+</script>
+
+<style lang="scss">
```
CarlSchwan, line 86: scoped?

src/components/VirtualScrolling.vue
```diff
@@ -0,0 +1,246 @@
...
+
+	methods: {
+		updateScrollPosition() {
+			this.scrollPosition = this.$refs.container.scrollTop
+		},
+	},
+}
+</script>
+
+<style lang="scss">
```
CarlSchwan, line 237: scoped?

src/components/TiledLayout.vue
artonge: Nice catch thanks :)

src/components/File.vue
artonge: Unsure what the solution is. Something like that ?

```suggestion
		<CheckboxRadioSwitch v-if="allowSelection"
			class="selection-checkbox"
			:checked="selected"
			@update:checked="onToggle">
			<span style="visibility: hidden">{{ t('photos', 'Select image {imageName}', {imageName: item.basename}) }}</span>
		</CheckboxRadioSwitch>
```

artonge: I used the above solution, so marking as resolve, feel free to ping me if you were thinking of something else :)

CarlSchwan: <this should use a `<label>` with a `for="<id of the checkboxradio`"` or a `aria-labelledby=""` (but make sure it also appear in the final `<input>` element as I'm not sure that `CheckboxRadioSwitch` propagate the `$args`)

artonge: Can't it with your suggestion, as `CheckboxRadioSwitch` does not propagate `$args` and the `id` is unknown. Above solution works :).

CarlSchwan: it doesn't really work since the screen reader needs the label to map a checkbox to a description, otherwise the checkbox and description are not linked

Probably will need additional API in nextcloud-vue

artonge: The content passed to `CheckboxRadioSwitch` is inside a `<label>` isn't it enought ?

skjnldsv: Should be enough yes, the `CheckboxRadioSwitch` is already accessible

skjnldsv:
🎉🎉🎉🎉🎉🎉

PR #1142, opened 2022-07-27
Add Albums view

artonge:
:heavy_check_mark:  **Done and ready for reviews** :heavy_check_mark: 

## Side quests

- [ ] Long press on mobile ?
- [ ] Use exif data to sort files
- [ ] Bug in `DAV:SEARCH` sorted by creationdate (some files from page 2 are before some files of page 1)
- [ ] Issue with `nc-vue::Action` when parent is sticky
- [ ] Sharing
    - [ ] Internal and link sharing of albums should be possible to make it on par with folders
    - [ ] See mockups, 
    - [ ] Uploading via link would be nice
    - [ ] Federation can be left out for now
    - [ ] When you add a photo, or others upload a photo, they both land in "Photos" – that’s the default, and it’s configurable

## Screenshots

Empty album list | Album list with content
-- | --
Screenshot 2022-08-11 at 12-05-14 Photos - Nextcloud | Screenshot from 2022-08-11 12-04-25

Album form modal | Album form with a name filled | Album form loading
-- | -- | --
Screenshot from 2022-08-11 12-05-36 | Screenshot from 2022-08-11 12-05-56 | Screenshot from 2022-08-11 12-06-12

Empty album content | Album content with content | Renaming an album
-- | -- | --
Screenshot 2022-08-11 at 11-44-05 Photos - Nextcloud | Screenshot 2022-08-11 at 11-38-16 Photos - Nextcloud | Screenshot 2022-08-11 at 11-39-11 Photos - Nextcloud

File picker preloading | File picker with content | File picker loading
-- | -- | --
Screenshot from 2022-08-11 11-47-41 | Screenshot 2022-08-11 at 11-39-38 Photos - Nextcloud | Screenshot 2022-08-11 at 12-13-56 Photos - Nextcloud

Album picker | Album picker preloading
-- | --
Screenshot 2022-08-11 at 12-16-52 Photos - Nextcloud | Screenshot from 2022-08-11 11-47-21

PR #1146, opened 2022-08-02
Album dav api

icewind1991:
Allow creating and managing folders trough dav

- create albums with `mkcol`
  `curl -u admin:admin -X MKCOL http://172.19.0.2/remote.php/dav/photos/admin/albums/test`
- add photos using `COPY`
  `curl -u admin:admin -H 'Destination: http://172.19.0.2/remote.php/dav/photos/admin/albums/test2/2.png' -X COPY http://172.19.0.2/remote.php/dav/files/admin/background.png'`
  note that the destination name is ignored
- rename albums using `MOVE`
  `curl -u admin:admin -H 'Destination: http://172.19.0.2/remote.php/dav/photos/admin/albums/test' -X MOVE 'http://172.19.0.2`/remote.php/dav/photos/admin/albums/test2'
- list albums or albums contents with PROPFIND (you can use `depth: 2` to list all albums with content included)
- file names in albums are prefixed by fileid to avoid collisions, you can request the `{http://nextcloud.org/ns}file-name` property to get the original filename
- full file can be downloaded from the album dav endpoint, thumbnails can be retrieved from the normal preview api by fileid

### DIFF
--- a/lib/Album/AlbumInfo.php
+++ b/lib/Album/AlbumInfo.php
@@ -1,0 +1,48 @@
+<?php
+
+declare(strict_types=1);
+/**
+ * @copyright Copyright (c) 2022 Robin Appelman <robin@icewind.nl>
+ *
+ * @license GNU AGPL version 3 or any later version
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+namespace OCA\Photos\Album;
+
+class AlbumInfo {
+	private int $id;
+	private string $userId;
+	private string $title;
+
+	public function __construct(int $id, string $userId, string $title) {
+		$this->id = $id;
+		$this->userId = $userId;
+		$this->title = $title;
+	}
+
+	public function getId(): int {
+		return $this->id;
+	}
+
+	public function getUserId(): string {
+		return $this->userId;
+	}
+
+	public function getTitle(): string {
+		return $this->title;
+	}
+}
--- a/lib/Album/AlbumInfo.php
+++ b/lib/Album/AlbumInfo.php
@@ -27,11 +27,24 @@ class AlbumInfo {
 	private int $id;
 	private string $userId;
 	private string $title;
+	private string $location;
+	private int $created;
+	private int $lastAdded;
 
-	public function __construct(int $id, string $userId, string $title) {
+	public function __construct(
+		int $id,
+		string $userId,
+		string $title,
+		string $location,
+		int $created,
+		int $lastAdded
+	) {
 		$this->id = $id;
 		$this->userId = $userId;
 		$this->title = $title;
+		$this->location = $location;
+		$this->created = $created;
+		$this->lastAdded = $lastAdded;
 	}
 
 	public function getId(): int {
@@ -44,5 +57,17 @@ class AlbumInfo {
 
 	public function getTitle(): string {
 		return $this->title;
+	}
+
+	public function getLocation(): string {
+		return $this->location;
+	}
+
+	public function getCreated(): int {
+		return $this->created;
+	}
+
+	public function getLastAddedPhoto(): int {
+		return $this->lastAdded;
 	}
 }
--- a/lib/Album/AlbumMapper.php
+++ b/lib/Album/AlbumMapper.php
@@ -1,0 +1,157 @@
+<?php
+
+declare(strict_types=1);
+/**
+ * @copyright Copyright (c) 2022 Robin Appelman <robin@icewind.nl>
+ *
+ * @license GNU AGPL version 3 or any later version
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+namespace OCA\Photos\Album;
+
+use OCP\DB\QueryBuilder\IQueryBuilder;
+use OCP\Files\IMimeTypeLoader;
+use OCP\IDBConnection;
+
+class AlbumMapper {
+	private IDBConnection $connection;
+	private IMimeTypeLoader $mimeTypeLoader;
+
+	public function __construct(IDBConnection $connection, IMimeTypeLoader $mimeTypeLoader) {
+		$this->connection = $connection;
+		$this->mimeTypeLoader = $mimeTypeLoader;
+	}
+
+	public function create(string $userId, string $name): AlbumInfo {
+		$query = $this->connection->getQueryBuilder();
+		$query->insert("photos_albums")
+			->values([
+				'user' => $query->createNamedParameter($userId),
+				'name' => $query->createNamedParameter($name),
+			]);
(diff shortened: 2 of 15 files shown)


lib/Album/AlbumMapper.php
```diff
@@ -0,0 +1,157 @@
...
+
+		$result = [];
+		foreach ($albumsById as $id => $album) {
+			$result[] = new AlbumWithFiles($album, $filesByAlbum[$id] ?? []);
+		}
+		return $result;
+	}
+
+	public function addFile(int $albumId, int $fileId): void {
+		$query = $this->connection->getQueryBuilder();
```
artonge, line 141: Check if file is already in the album ?

lib/Migration/Version20000Date20220727125801.php
```diff
@@ -0,0 +1,85 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * @copyright Copyright (c) 2022 Your name <your@email.com>
```
artonge, line 6: Copyright is wrong

lib/Sabre/Album/AlbumPhoto.php
```diff
@@ -0,0 +1,99 @@
...
+		$this->file = $file;
+		$this->userFolder = $userFolder;
+	}
+
+	public function delete() {
+		$this->albumMapper->removeFile($this->album->getId(), $this->file->getFileId());
+	}
+
+	public function getName() {
+		return $this->file->getFileId() . "-" . $this->file->getName();
```
artonge, line 54: Why include the file id ?

lib/Sabre/Album/AlbumRoot.php
```diff
@@ -0,0 +1,117 @@
+<?php
+
+declare(strict_types=1);
+/**
+ * @copyright Copyright (c) 2021 Robin Appelman <robin@icewind.nl>
```
artonge, line 5: 2022

lib/Sabre/Album/AlbumPhoto.php
artonge: Oh, to not have two files with the same name in an album ?

lib/Sabre/Album/AlbumRoot.php
```diff
@@ -0,0 +1,117 @@
...
+
+	public function getChildren(): array {
+		return array_map(function (AlbumFile $file) {
+			return new AlbumPhoto($this->albumMapper, $this->album->getAlbum(), $file, $this->userFolder);
+		}, $this->album->getFiles());
+	}
+
+	public function getChild($name): AlbumPhoto {
+		foreach ($this->album->getFiles() as $file) {
+			if ($file->getFileId() . "-" . $file->getName() === $name) {
```
artonge, line 83: Wouldn't it be better to compare fileId only ?

artonge (changes requested):
I don't see any way to store the album location. Also, there is nothing about the sharing state (collaborators).

Will it come in future work ?

lib/Sabre/Album/AlbumPhoto.php
icewind1991: Yes

lib/Sabre/Album/AlbumsHome.php
```diff
@@ -0,0 +1,108 @@
...
+	}
+
+	public function createDirectory($name) {
+		$uid = $this->user->getUID();
+		$this->albumMapper->create($uid, $name);
+	}
+
+	public function getChild($name) {
+		foreach ($this->getChildren() as $child) {
+			if ($child->getName() === $name) {
```
artonge, line 78: I am right if I say that two albums will not be able to have the same name ? It makes sense in a DAV API, but for the user it might be confusing. Maybe we should distinguish between name and display name. But thinking about it I am not so sure.

@jancborchardt or @nimishavijay, should the user be able to have two albums with the same name ?

lib/Album/AlbumMapper.php
icewind1991: There is a check in the dav api

icewind1991:
> I don't see any way to store the album location

How do we want to store location? coordinates or freeform string?

artonge:
> How do we want to store location? coordinates or freeform string?

Freeform string

lib/Album/AlbumMapper.php
```diff
@@ -0,0 +1,157 @@
...
+		foreach ($rows as $row) {
+			$albumId = $row['album_id'];
+			if ($row['fileid']) {
+				$mimeId = $row['mimetype'];
+				$mimeType = $this->mimeTypeLoader->getMimetypeById($mimeId);
+				$filesByAlbum[$albumId][] = new AlbumFile($row['fileid'], $row['file_name'], $mimeType, $row['size'], $row['mtime'], $row['etag']);
+			}
+
+			if (!isset($albumsById[$albumId])) {
+				$albumsById[$albumId] = new AlbumInfo($albumId, $userId, $row['album_name']);
```
artonge, line 129: From my testing: `$albumId` is a string, but int is expected.

artonge:
Also, could there be a way to have the creation date, the number of items, and a cover picture ? I can be the latest added picture for now. The last two information could be retrieved from fetching the file list, but it won't scale with a lot of albums.

lib/Sabre/Album/AlbumPhoto.php
```diff
@@ -0,0 +1,99 @@
...
+use OCA\Photos\Album\AlbumInfo;
+use OCA\Photos\Album\AlbumMapper;
+use OCP\Files\Folder;
+use OCP\Files\Node;
+use OCP\Files\File;
+use OCP\Files\NotFoundException;
+use Sabre\DAV\Exception\Forbidden;
+use Sabre\DAV\IFile;
+
+class AlbumPhoto implements IFile {
```
artonge, line 36: Could we add the fileid to the returned data ?

artonge (changes requested):

lib/Sabre/Album/AlbumPhoto.php
artonge, line 36: Is it possible to request the file's metadata ?

artonge (changes requested):

lib/Sabre/Album/AlbumsHome.php
PVince81: I'd say no as it makes things more complicated (can of worms) for not much benefit

lib/Album/AlbumMapper.php
artonge: Happens in other places

artonge:
Additional work done here: https://github.com/nextcloud/photos/pull/1142

lib/Album/AlbumMapper.php
icewind1991: should be fixed

lib/Sabre/Album/AlbumPhoto.php
icewind1991: done

icewind1991: done

icewind1991:
Added `location` and `last-photo` to albums at `{http://nextcloud.org/ns}location` and `'{http://nextcloud.org/ns}last-photo`, the location can be set through `PROPPATCH`

Note that this change re-uses the same migration which will not run automatically if you've already run a previous version of it. You can manually trigger the new changes by deleting the `photos_albums` and `photos_albums_files` tables and running `occ migrations:execute photos 20000Date20220727125801`

matiasdelellis:
First of all, excuse the boldness.
I understand that it is intended for user albums, but can it also be extended as virtual albums filled by applications?.

In principle, perhaps it would only need a flag to indicate that it cannot be edited by the user [1], perhaps change 'location' to 'context' [2], and an 'icon' would be interesting to help the context.

[1]: Ideally, it would be necessary to make an extensive api (Registration of applications, dynamic responses, hooks/events for removal/add photos, similar to searches), but I understand that it would already be beyond the intention of this proposal, and indicating that it cannot be edited and that it is generated automatically may be enough.

[2]: Just a name change, since in an album you don't always organize by places, and it can be used to comment who generates the album. 🤔 

[3]: If it's not by location, it may be by the content of the image (recognize, facerecognition, maps apps), and a simple icon would be fine.

Regards,
Matias

lib/Album/AlbumInfo.php
```diff
@@ -0,0 +1,73 @@
...
+ */
+
+namespace OCA\Photos\Album;
+
+class AlbumInfo {
+	private int $id;
+	private string $userId;
+	private string $title;
+	private string $location;
+	private int $created;
```
artonge, line 31: Not needed anymore. Album's date is now dynamic based on its content: `oldest item's month and year - newest item's month and year`
https://github.com/nextcloud/photos/pull/1142/files#diff-9706d467ff74e216f41e0da8dd6424754754a63018ff15b3e7fadcef9d33f811R118-R134

artonge (approved):

marcelklehr:
@matiasdelellis Thank you for your input! We're iterating on this, currently, this is the scope we're working on. It may be that virtual albums will appear in the future as a refactoring or similar. Right now we're being pragmatic :)

PR #1162, opened 2022-08-24
Refactor: Use a global AbortController

marcelklehr:
The rationale for this is that request code currently lives in components without any way to keep the code DRY.
Ideally, request code would live in vuex actions IMO, but then we lose the ability to cancel requests from views.
This PR removes logic from views that handles request cancellation and moves it to the router, which cancels requests upon route changes. ~~I've also introduced a global priority queue for requests to replace the semaphore business, which also would have to be added to every request otherwise.~~


src/store/albums.js
```diff
@@ -266,7 +267,7 @@ const actions = {
 		try {
 			context.commit('updateAlbum', { album: updatedAlbum })
 
-			await client.customRequest(
+			await RequestHandler.queue.add(() => client.customRequest(
```
artonge, line 270: Should not be aborted on route change.

src/views/AlbumContent.vue
```diff
@@ -309,24 +309,17 @@ export default {
 				return []
 			}
 
-			const semaphoreSymbol = await this.semaphore.acquire(() => 0, 'fetchFiles')
-			const fetchSemaphoreSymbol = await this.fetchSemaphore.acquire()
```
artonge, line 313: Semaphores are used to prioritize XHR requests compared to photos requests. Removing it will disable this functionality

artonge (changes requested):
Overall, I think only requests that fetch photos or albums should be canceled. They are read only requests and the only ones that can take a long time to load. The other ones are here to alter the DB, so they should not be canceled.

I would prefer for the views to keep the responsibility to cancel the requests, and not a central service. Using a mixin here might make more sense.

src/views/AlbumContent.vue
marcelklehr: The PQueue has a priority argument, which could be used for this. Then we wouldn't have prioritization littered over the code base. What do you think about that?

artonge: Priority of potos loading might change as the user scrolls.

If we want to prioritize xhr over photos, and have dynamic priority for photos, then semaphore is a good fit

src/store/albums.js
marcelklehr: Fixed

src/views/AlbumContent.vue
marcelklehr: I see. I've removed my changes to the semaphore code

src/router/index.js
```diff
@@ -179,3 +180,10 @@ export default new Router({
 		},
 	],
 })
+
+router.beforeResolve((from, to, next) => {
+	cancelAll()
+	next()
+})
```
artonge, line 187: We lose aborting request on destroy, useful for components like the `FilePicker`. Wouldn't a mixin be more flexible here ?

marcelklehr: do you mean putting only the cancelAll() call into a mixin, so views can decide whether they want to use it?

artonge: And the `abortController` creation, exposing it on `data`.

Something like:


```js
export default {
	name: 'CancelRequestMixin',

	data() {
		return {
			abortController: new AbortController()
		}
	},

	beforeDestroy() {
		this.abortController.abort()
	},

	beforeRouteLeave() {
		this.abortController.abort()
	},
}
```

marcelklehr: But then how do we get the abort signal to vuex actions? Via params?

artonge: Yes, the same way you did it on `Tag.vue` for example:


```js
			const files = await getTaggedImages(this.item.injected.id, {
				signal: this.abortController.signal,
			})
```

or

```js
			const files = await getTaggedImages(this.item.injected.id, {
				signal: this.abortSignal,
			})
```

artonge (approved):

PR #1184, opened 2022-08-26
Bump webdav to ^4

skjnldsv:
Removed unused service since 3 years...
https://github.com/nextcloud/photos/pull/4/files#diff-239342c3890dc7339f0c87a00179a78c0502d3f74b5b387f18fd2e00cb19d194L57

Webdav 3 is deprecated and don't receive updates anymore


marcelklehr:
Does the client API stay the same? That's nice!

skjnldsv:
> Does the client API stay the same? That's nice!

Yes and no, I had to adapt a few hacks we do as well as their new esm export.

webpack.js
```diff
@@ -41,9 +42,10 @@ webpackConfig.module.rules = Object.values(webpackRules)
 webpackConfig.plugins.push(
 	// patch webdav/dist/request.js
 	new webpack.NormalModuleReplacementPlugin(
-		/request.js/,
+		/request(\.js)?/,
```
marcelklehr, line 45: oof, I didn't know about this. :'D

marcelklehr (approved):

webpack.js
skjnldsv: We could make it upstream, but https://github.com/perry-mitchell/webdav-client/issues/179

webpack.js
```diff
@@ -41,9 +42,10 @@ webpackConfig.module.rules = Object.values(webpackRules)
...
 		function(resource) {
-			if (resource.context.indexOf('webdav/dist') > -1) {
+			if (resource.context.indexOf('webdav') > -1) {
+				console.debug('Patched request for webdav', basename(resource.contextInfo.issuer))
```
marcelklehr, line 48: This doesn't seem to work on latest master, though :S I've put a direct throw Error into patchedRequest and it never gets thrown?

PR #1189, opened 2022-08-30
Implement default Photos folder

skjnldsv:
image

- [x] Move settings to proper Modal
- [x]  Added default Photos location
   Default set to `/Photos` https://github.com/nextcloud/photos/blob/a07c06ff6919ddba691eb1ee110d997127d1975b/lib/Service/UserConfigService.php#L37
- [x]  Implement UserConfig service
- [x]  Cleanup code


lib/Service/UserConfigService.php
```diff
@@ -0,0 +1,73 @@
...
+namespace OCA\Photos\Service;
+
+use Exception;
+use OCA\Photos\AppInfo\Application;
+use OCP\Files\IRootFolder;
+use OCP\IConfig;
+use OCP\IUserSession;
+
+class UserConfigService {
+	public const DEFAULT_CONFIGS = [
```
skjnldsv, line 35: Could theoritecally be exposed as a capability and used straight away by the https://github.com/nextcloud/photos/blob/34948e3f8ad93a30a68cdc0adebf91f642d141ff/src/mixins/UserConfig.js#L35 instead of having to declare them again in the front :)

src/components/Settings/PhotosLocationSettings.vue
```diff
@@ -0,0 +1,89 @@
...
+				return
+			}
+
+			if (path.includes('//')) {
+				path = path.replace(/\/\//gi, '/')
+			}
+
+			this.photosLocation = path
+			this.updateSetting('photosLocation')
+		},
```
artonge, line 86: Or:

```suggestion
		updatePhotosFolder: debounce(function (path) {
			console.debug(`Path '${path}' selected for photos location`)
			if (typeof path !== 'string' || path.trim() === '' || !path.startsWith('/')) {
				showError(t('photos', 'Invalid location selected'))
				return
			}
			if (path.includes('//')) {
				path = path.replace(/\/\//gi, '/')
			}
			this.photosLocation = path
			this.updateSetting('photosLocation')
		}),
```

But works as you coded it so feel free to ignore.

src/components/Settings/PhotosLocationSettings.vue
```diff
@@ -0,0 +1,89 @@
...
+				.startAt(this.photosLocation)
+				.build()
+
+			picker.pick()
+				.then(this.updatePhotosFolder)
+		},
+
+		debounceUpdatePhotosFolder: debounce(function(...args) {
+			this.updatePhotosFolder(...args)
+		}, 300),
```
artonge, line 71: Is this not working?

```suggestion
		debounceUpdatePhotosFolder: debounce(this.updatePhotosFolder, 300),
```

skjnldsv: let me check :)

src/components/Settings/PhotosLocationSettings.vue
skjnldsv: debounce requires a function reference, here the function will be created on each call, so it will be different and not debounced properly, no?

src/components/Settings/PhotosLocationSettings.vue
skjnldsv: No, because `this` is not set there otherwise

src/components/Settings/PhotosLocationSettings.vue
skjnldsv: Also, sometimes you don't need to debounce, like right after picking with the Folder picker for example

artonge: > debounce requires a function reference, here the function will be created on each call, so it will be different and not debounced properly, no?

Here, debounce will be called only once at component creation and return a debounced function from the passed one. So the function will stay the same

artonge (approved):

PR #1191, opened 2022-08-30
Rework tags view

marcelklehr:
based on #1162

Screencast: https://cloud.marcelklehr.de/s/sxEm6Nd65xT8GZW/download/Peek%202022-08-31%2012-22.gif


jancborchardt (changes requested):
Wow so cool! Looks really nice :) Some details:
- [ ] The popular tags are not sorted by ones with most photos yet
- [ ] Some of the previews seem to have square corners instead of rounded, possibly cause they are not filling the container, not sure?
- [ ] The titles "Popular tags" and "All tags" could be moved a bit to the right to left-align with the leftmost image preview (not the hover effect)
- [ ] The albums look quite large, or is that just me? Maybe it’s fine. :) Is the size fixed, or calculated based on screen size? (cc @artonge as I guess it just uses the album size and we should make sure the sizing is the same :)

Otherwise all really really nice!

artonge:
> [ ] The albums look quite large, or is that just me? Maybe it’s fine. :) Is the size fixed, or calculated based on screen size? (cc @artonge as I guess it just uses the album size and we should make sure the sizing is the same :)

Not the same component, but I assume @marcelklehr used the same size.

nimishavijay:
Woah looks super cool! :) Agree with everything, especially the sizing of the photos, plus possibly: 
- when you are viewing the photos of a tag, show the number of photos as a subline to the title of the tag

marcelklehr:
> The albums look quite large, or is that just me?

Tag cover size is the same as album covers, but will get smaller on mobile.

> The popular tags are not sorted by ones with most photos yet

Done.

marcelklehr:
image

jancborchardt (approved):
Looks great, good to go design-wise! :)

artonge (approved):
Looks good.

Doesn't it miss some padding in the TagContent. Is it en purpose ?

marcelklehr:
> Doesn't it miss some padding in the TagContent. Is it en purpose ?

I didn't add any padding after reworking, yeah. Do we want padding on the photos container @nimishavijay @jancborchardt?

jancborchardt:
> I didn't add any padding after reworking, yeah. The timeline view for example indeed has padding. Do we want padding on the photos container here @nimishavijay @jancborchardt?

Right, yes would be good if it has the same padding as on the timeline view. :)

PR #1194, opened 2022-08-31
Implement upload in Folder

skjnldsv:
https://user-images.githubusercontent.com/14975046/187891817-048e699e-14c6-422d-98b6-21937b639066.mp4


skjnldsv:
@artonge ready for a first review! If you have any questions :)
Currently only file upload in folders.
I did not use the default folder as if you're displaying the root and upload a file, you would expect this to be in the Root folder I think

src/views/Folders.vue
```diff
@@ -29,14 +29,24 @@
...
-			key="navigation"
-			v-bind="folder"
+	<div v-else-if="!initializing">
+		<HeaderNavigation key="navigation"
+			:loading="loading"
+			:path="folder.filename"
+			:title="folder.basename"
 			:root-title="rootTitle"
-			:show-actions="true" />
+			@refresh="onRefresh">
```
skjnldsv, line 43: Clicking the title refreshes the view

src/views/Folders.vue
```diff
@@ -29,14 +29,24 @@
...
+			<UploadPicker :accept="allowedMimes"
+				:destination="folder.filename"
+				:multiple="true"
+				@uploaded="onRefresh" />
```
skjnldsv, line 47: Every uploaded file request a refresh. For big folders, that could be expensive, maybe we could fetch the newly uploaded file only? Not sure the albums API allows it :/

artonge: This is to progressively display the files will they are uploaded ?

src/components/FolderTagPreview.vue
```diff
@@ -65,7 +65,7 @@ export default {
 			default: 'icon-folder',
 		},
 		id: {
-			type: Number,
+			type: [Number, String],
```
marcelklehr, line 68: Whaat, I didn't know that was a thing. Nice!

src/views/Folders.vue
marcelklehr: Shouldn't it be enough to add the file id to the folder list without refreshing the full view?

skjnldsv: We don't know the File id without fetching the content :see_no_evil: 

> This is to progressively display the files will they are uploaded ?

This or update the view after upload

artonge: Might be best to do a propfind on the new file, this way we are sure to have all the data, and less expensive than propfinding all the list. This is what files is doing apparently.

artonge (approved):

PR #1204, opened 2022-09-03
fix service-worker

szaimen:
Found via https://github.com/search?q=org%3Anextcloud+noRewrite&type=code

Disclaimer: I wasn't able to test this but I think it should work.

See also https://github.com/nextcloud/bookmarks/issues/1078#issuecomment-1236174504

Signed-off-by: szaimen <szaimen@e.mail.de>


src/Photos.vue
```diff
@@ -165,7 +165,11 @@ export default {
 		if ('serviceWorker' in navigator) {
 			// Use the window load event to keep the page load performant
 			window.addEventListener('load', () => {
-				navigator.serviceWorker.register(generateUrl('/apps/photos/service-worker.js'), {
+				const webRootWithIndexPHP = getRootUrl() + '/index.php'
+				const doesURLContainIndexPHP = window.location.pathname.startsWith(webRootWithIndexPHP)
+				navigator.serviceWorker.register(generateUrl('/apps/photos/service-worker.js', {}, {
+					noRewrite: doesURLContainIndexPHP,
```
szaimen, line 171:
```suggestion
					noRewrite: !doesURLContainIndexPHP,
```
we may need to invert the logic here to make it work. not sure though...

szaimen: or we simply always set it to true, WDYT?

szaimen: done

artonge (approved):

skjnldsv (changes requested):
Let's wait for #1206

skjnldsv (approved):

PR #1205, opened 2022-09-05
Fix scrolling for new design

artonge:


marcelklehr:
TagContent and floders views are also broken. Not sure about On this day, favorites, people and shared with you

src/views/AlbumContent.vue
```diff
@@ -530,14 +531,13 @@ export default {
 	}
 
 	&__photos {
-		height: 100%;
+		height: calc(100% - 60px);
```
marcelklehr, line 534: This makes FileListViewer the scrollable element, but shouldn't the app content be the scrollable element?

marcelklehr: Ah, we probably need vue v7 for that, right?

artonge: At this point, we don't really care which element is scrollable, no ?

artonge:
Superseded by https://github.com/nextcloud/photos/pull/1206

PR #1206, opened 2022-09-05
Bump @nextcloud/vue to v7

artonge:


skjnldsv (approved):
Pushed a fix.
I removed the usage of `NcTextField` because it have a bug on the focus event it seems (focusing the input triggers the focus event more than one time for god know why. So let's keep it simple for now)

Albums have an horizontal scrollbar and a weird rendering data issue
image

We can fix it in a followup if you want

src/components/EmptyContent.vue
```diff
@@ -1,140 +0,0 @@
...
- -
- - You should have received a copy of the GNU Affero General Public License
- - along with this program. If not, see <http://www.gnu.org/licenses/>.
- -
- -->
-
-<template>
-	<div v-if="isReady" class="emptycontent">
-		<!-- eslint-disable-next-line vue/no-v-html (because it's an SVG file) -->
-		<div v-if="haveIllustration" class="illustration" v-html="illustration" />
```
skjnldsv, line 26: @artonge regression, we were not using the vue component.
So you, migrating to it, broke support for illustrations as well as default empty contents :)
https://github.com/nextcloud/photos/blob/ce56825aa759b6ba8d5175579af75818cceeea11/src/views/Tags.vue#L27-L29
https://github.com/nextcloud/photos/blob/ce56825aa759b6ba8d5175579af75818cceeea11/src/views/Tags.vue#L31-L36

artonge: Seen, and on it ;) trying to abstract those views to have more consistency between them

src/views/TagContent.vue
```diff
@@ -24,34 +24,34 @@
...
 		{{ t('photos', 'An error occurred') }}
-	</EmptyContent>
+	</NcEmptyContent>
 
 	<!-- Folder content -->
 	<div v-else-if="!loading">
 		<div class="photos-navigation">
-			<Actions class="photos-navigation__back">
+			<NcActions class="photos-navigation__back">
 				<ActionButton @click="$router.push({name: 'tags'})">
```
marcelklehr, line 35: Regression. I'm on it.

PR #1207, opened 2022-09-06
Fix viewer editor permissions

### LINKED ISSUE #315: Delete command from sidebar list
Missing a easy fast DELETE command. Maybe in the side bar menu ...

skjnldsv:
image

Fix #315


src/services/DavRequest.js
```diff
@@ -20,17 +20,19 @@
...
+	<d:resourcetype />
 	<nc:face-detections />
+	<nc:file-metadata-size />
+	<nc:has-preview />
+	<nc:realpath />
 	<oc:favorite />
-	<d:resourcetype />`
+	<oc:fileid />
+	<oc:permissions />
+`
```
artonge, line 35: Any reason for the reordering ? just curious :)

artonge (approved):

src/services/DavRequest.js
marcelklehr: just to mess with us :D

marcelklehr:
Would this be needed for the tags and people views as well?

artonge:
@skjnldsv ^

src/services/DavRequest.js
skjnldsv: Sorting alphabetically

skjnldsv:
> Would this be needed for the tags and people views as well?

Only if you open media on Viewer from those views :shrug:

marcelklehr:
> Only if you open media on Viewer from those views

that's probably a yes...

lib/Controller/AlbumsController.php
```diff
@@ -109,13 +110,36 @@ private function formatData(iterable $nodes): array {
 				'mime' => $node->getMimetype(),
 				'size' => $node->getSize(),
 				'type' => $node->getType(),
+				'permissions' => $this->formatPermissions($node->getPermissions()),
```
marcelklehr, line 113: Ah, and this is not the albums controller but the folders controller, probably, then we may need this for the actual albums as well

skjnldsv: Yep, but the other Views are importing `src/services/DavRequest.js`, so they will benefit from that too

skjnldsv: https://github.com/nextcloud/photos/search?q=DavRequest

marcelklehr: Yes, but Albums and Faces dav endpoints do not expose permissions in DAV afaik

PR #1208, opened 2022-09-06
Recreate an abort controller on route leave

artonge:
This is so new requests can be made by the component


skjnldsv (approved):

marcelklehr:
What if new requests are made while the route is changing, won't these new requests use the newly created abort controller then? Instead of being auto-cancelled? Would it make sense to add the recreation code to a routeEnter hook?

artonge:
> What if new requests are made while the route is changing, won't these new requests use the newly created abort controller then? Instead of being auto-cancelled? Would it make sense to add the recreation code to a routeEnter hook?

Unsure how this could happen unless there is some code inside a `beforeRouteLeave` hook ?
The `beforeRouteEnter` hook is probably called just after `beforeRouteLeave`, so the issue would be the same for other cases.
Also, `beforeRouteEnter` is probable called on initial load, and it does not make sens to recreate the abort controller at that point.

And I think that keeping the recreation logic just after we call `abort()` make more sense. We recreate one just after we burned the current one.

PR #1209, opened 2022-09-06
Reduce the number of loaded media

artonge:
It put too much pressure on the server RAM


skjnldsv (approved):

artonge:
Well, looks like it was included in https://github.com/nextcloud/photos/pull/1208
Black magic is the only explanation

marcelklehr:
git thought it's a good idea and slipped it in :D

PR #1225, opened 2022-09-08
Allow to PUT files into an album

skjnldsv:
First step towards Albums upload
- If a PUT is caught, we request the default Photos user folder and we upload it there
- Then we create the ink

### DIFF
--- a/lib/Sabre/Album/AlbumsHome.php
+++ b/lib/Sabre/Album/AlbumsHome.php
@@ -25,6 +25,7 @@ namespace OCA\Photos\Sabre\Album;
 
 use OCA\Photos\Album\AlbumMapper;
 use OCA\Photos\Album\AlbumWithFiles;
+use OCA\Photos\Service\UserConfigService;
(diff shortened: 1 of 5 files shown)


lib/Sabre/Album/AlbumsHome.php
```diff
@@ -104,7 +107,7 @@ public function getChildren(): array {
 		if ($this->children === null) {
 			$folders = $this->albumMapper->getForUserWithFiles($this->user->getUID());
 			$this->children = array_map(function (AlbumWithFiles $folder) {
-				return new AlbumRoot($this->albumMapper, $folder, $this->rootFolder, $this->userFolder, $this->user);
+				return new AlbumRoot($this->albumMapper, $folder, $this->rootFolder, $this->userFolder, $this->user, $this->userConfigService);
```
skjnldsv, line 110: All of this is super cumbersome btw @icewind1991 
Why can't we use DI here?

artonge (approved):

PR #1240, opened 2022-09-14
FaceCovers: Don't use pictures with the face at the edge

marcelklehr:
|before|after|
|-|-|
|Screenshot from 2022-09-14 13-38-43|Screenshot from 2022-09-14 13-30-14|

I couldn't get it to put the faces at the same position due to a different bug.


artonge (approved):

jancborchardt (approved):
Nice, looks great! :)

PR #1241, opened 2022-09-14
FaceCovers: Hide first description line if there's no face title

marcelklehr:
image


jancborchardt:
Could you show how it looks like with 1 person actually having a name set? It would be good if the text baseline is the same, no matter if it’s a name or a "x photos" line.

(I am thinking that probably the "x photos" lines are a bit further up cause they are smaller, so would be nice to align them.)

marcelklehr:
image

jancborchardt (approved):
Looks great, thank you! :)

PR #1243, opened 2022-09-16
Add public sharing front-end

artonge:


PVince81 (approved):
:+1: for now, backend implementation will follow later

Issue #1260, opened 2022-09-19
Cannot add a NOT NULL column with default value NULL


blizzz:
**Describe the bug**

An upgrade from latest 24 fails due to an Exception while Migration processing. May also happen on fresh installations.

**To Reproduce**
1. Upgrade from latest NC to latest build https://download.nextcloud.com/server/prereleases/nextcloud-25.0.0beta7.zip (not officially released)

**Expected behavior**
Upgrades runs through successfully

**Logs**

```
Updating <photos> ...
Exception: Database error when running migration 20001Date20220830131446 for app photos
An exception occurred while executing a query: SQLSTATE[HY000]: General error: 1 Cannot add a NOT NULL column with default value NULL
Update failed
Maintenance mode is kept active
Resetting log level
    0/0 [--->------------------------]   0%
Error: Process completed with exit code 5.
```

from this automated build run https://github.com/nextcloud/updater_server/actions/runs/3083044439/jobs/4983518734

nickvergessen:
Setting
https://github.com/nextcloud/photos/blob/master/lib/Migration/Version20001Date20220830131446.php#L72
to false should be enough, otherwise you need to specify a default value.

PVince81:
a bit tricky, I'd expect "owner" to be always populated but we can't do during the migration

maybe for now we allow it to be null and @artonge can figure out a way to populate it or use a default value later

PVince81:
PR for unblocking: https://github.com/nextcloud/photos/pull/1261

let's leave the issue open for further fixing

PR #1261, opened 2022-09-19
Make photos_albums_files.owner nullable for beta migrations

PVince81:
For https://github.com/nextcloud/photos/issues/1260 to unblock beta7

### DIFF
--- a/lib/Migration/Version20001Date20220830131446.php
+++ b/lib/Migration/Version20001Date20220830131446.php
@@ -69,7 +69,7 @@ class Version20001Date20220830131446 extends SimpleMigrationStep {
 			$modified = true;
 			$table = $schema->getTable("photos_albums_files");
 			$table->addColumn('owner', Types::STRING, [
-				'notnull' => true,
+				'notnull' => false,
 				'length' => 64,
 			]);
 		}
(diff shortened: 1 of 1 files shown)


nickvergessen (approved):

PVince81:
or we ask users to purge that table if updating from another beta...

the table was indeed added recently in beta3: https://github.com/nextcloud/photos/commit/10f44aacc8f5161655694fff2dcca0f8d83ba3d1

lib/Migration/Version20001Date20220830131446.php
```diff
@@ -69,7 +69,7 @@ public function changeSchema(IOutput $output, Closure $schemaClosure, array $opt
 			$modified = true;
 			$table = $schema->getTable("photos_albums_files");
 			$table->addColumn('owner', Types::STRING, [
-				'notnull' => true,
+				'notnull' => false,
```
blizzz, line 72:
```suggestion
				'notnull' => true,
				'default' => '',
```

blizzz (commented):
setting an emptry string as default instead and keeping the not-null constraint

PVince81:
yeah ok, let's go with empty string then @blizzz

blizzz:
> yeah ok, let's go with empty string then @blizzz

Sorry, that was shit advice

`Column "oc_photos_albums_files"."owner" is NotNull, but has empty string or null as default.`

lib/Migration/Version20001Date20220830131446.php
```diff
@@ -70,6 +70,7 @@ public function changeSchema(IOutput $output, Closure $schemaClosure, array $opt
 			$table = $schema->getTable("photos_albums_files");
 			$table->addColumn('owner', Types::STRING, [
 				'notnull' => true,
```
nickvergessen, line 72: Empty string is null in oracle 💥

nickvergessen (changes requested):

PR #1263, opened 2022-09-19
Add a github action that installs the app

nickvergessen:
Signed-off-by: Joas Schilling <coding@schilljs.com>


blizzz:
maybe adding the other DB's isn't that bad of an idea either? Either way, that certainly is already helpful!

nickvergessen:
The installer should check all db constraints already as oracle was not removed from supported dbs

Issue #1238, opened 2022-09-10
Albums optimisation


skjnldsv:
- [ ] Seems like this could get expensive pretty quickly. We should most likely store the date instead of parsing all files
  https://github.com/nextcloud/photos/blob/dbe72c433de0cc4b17972dd6d9a123c8c930b574/lib/Sabre/Album/AlbumRoot.php#L131-L147
- [ ] 
- [ ] 

cc @artonge

PR #1282, opened 2022-09-22
Do not get all the files list when fetching the albums list

skjnldsv:
Partial fix for https://github.com/nextcloud/photos/issues/1238

1. When propfind on the `/remote.php/dav/photos/admin/albums/`
    We used to fetch all albums WITH files. That meant fetching literally ALL the files in all albums for a specific user, so it will quickly get expensive
3. I moved the files cache on a per-album basis instead of on an album-root initialisation 

Advice welcome, we should also have psalm somewhere here! :see_no_evil: @CarlSchwan

### DIFF
--- a/lib/Album/AlbumMapper.php
+++ b/lib/Album/AlbumMapper.php
@@ -177,40 +177,33 @@ class AlbumMapper {
 	}
 
 	/**
+	 * @param int $albumId
 	 * @param string $userId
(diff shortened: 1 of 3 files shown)


artonge (approved):
Other endpoints have probably the same issue. Let's merge this one and see later for the others

PR #1294, opened 2022-09-26
update service worker for correct scope

szaimen:
Signed-off-by: szaimen <szaimen@e.mail.de>


src/Photos.vue
```diff
@@ -170,7 +170,7 @@ export default {
 				navigator.serviceWorker.register(generateUrl('/apps/photos/service-worker.js', {}, {
 					noRewrite: true,
 				}), {
-					scope: '/',
+					scope: generateUrl('/apps/photos'),
```
szaimen, line 173: However assets might be loaded through a different route so not sure if this would really work. Thoughts?

artonge (approved):

src/Photos.vue
artonge: Previews are now loaded from `/apps/photos/api/v1/preview` so it makes sense

artonge: From the doc: the `scope` property is the "range of URLs a service worker can control". So it makes sense.

https://developer.mozilla.org/en-US/docs/Web/API/ServiceWorkerContainer/register#parameters

szaimen: all right, cool!

skjnldsv: Initially we had an impact on all previews, but now that we are using a custom route for previews that only applies on photos, I guess that make sense :)

szaimen: Great! Is this custom route only for NC25?

skjnldsv: yep

skjnldsv: https://github.com/nextcloud/photos/blob/2cad0f85b7ccb4e98735823a8ca8b014d486029c/appinfo/routes.php#L123-L129

szaimen:
cypress failure unrelated

szaimen:
/backport https://github.com/nextcloud/photos/commit/51cdc208ed65880d3ab3f8b41e94ac5c4f2a221a to stable25

Issue #1297, opened 2022-09-27
NC25 RC1 changing default media location wont load path.


Yougoshatenshi:
**Describe the bug**
Photos App default media location wont load file path.

**To Reproduce**
1. Go to Photos Settings
2. Click on the file path

**Expected behavior**
Show path and able to select it.

**Screenshots**
![image](https://user-images.githubusercontent.com/13738830/192401242-206f58d0-45bc-4b7e-b003-c252c8df8baa.png)

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Browser: Mozilla Firefox
 - Version: 106.0b4

same with Microsoft Edge Version 105.0.1343.53 (Official build) (64-bit)

**Browser log**
TypeError: this.$filePicker is null

![image](https://user-images.githubusercontent.com/13738830/192401271-a7e6866c-1d49-40c1-bbd0-910acc1b5fc2.png)

skjnldsv:
Thanks for the find, fix in #1298

PR #1298, opened 2022-09-27
Fix file picker initialization

skjnldsv:
Fix https://github.com/nextcloud/photos/issues/1297

Seems like until we can properly reject the picker on close or ESC, we cannot really prevent a loop init.
The picker really needs to be re-written in vue. Until then, this should mitigate the problem


PVince81 (approved):
:+1:

blizzz:
master is 26, please trigger backport if necessary.

skjnldsv:
> master is 26, please trigger backport if necessary.

Thanks ! :)

PR #1300, opened 2022-09-27
Fix psalm errors

artonge:


szaimen (approved):
psalm test is green now

PR #1310, opened 2022-10-03
Add CLI and listeners to reverse geocode pictures' coordinates

artonge:
Add commands and listeners to generate location data of files:

The location data is stored inside `oc_files_metadata`.

## Added commands

- `occ photos:update-1000-cities` to update the cities1000 file.
- `occ photos:map-media-to-location`to map picture coordinates to a location

## Architecture

- `ReverseGeoCoderService` download the necessary files and build the `KDTree`
- `UpdateReverseGeocodingFilesCommand` command to allow to manually create the needed reverse geocoding files
- `MediaLocationManager` to manager the location mappings
- `MapMediaToLocationCommand` command to manually trigger location data mapping. Useful for pre-existing pictures.
- `LocationManagerNodeEventListener` to react to node, user and share events.
- `MapMediaToLocationJob` to reduce the load in event listeners

```php
                                          ┌─────────────────────┐
                            ┌────────────►│MapMediaToLocationJob│
                            │             └─────────┬───────────┘
                            │                       │
   ┌────────────────────────┴───────┐               │
   │LocationManagerNodeEventListener├──┐            ▼
   └────────────────────────────────┘  │  ┌────────────────────┐     ┌──────────────┐
                                       ├─►│MediaLocationManager├────►│LocationMapper│
   ┌─────────────────────────┐         │  └─────────┬──────────┘     └──────────────┘
   │MapMediaToLocationCommand├─────────┘            │
   └─────────────────────────┘                      │
                                                    ▼
   ┌──────────────────────────────────┐   ┌──────────────────────┐
   │UpdateReverseGeocodingFilesCommand├──►│ReverseGeoCoderService│
   └──────────────────────────────────┘   └──────────────────────┘
```

### DIFF
--- a/appinfo/info.xml
+++ b/appinfo/info.xml
@@ -1,13 +1,13 @@
 <?xml version="1.0"?>
-<info xmlns:xsi= "http://www.w3.org/2001/XMLSchema-instance"
-      xsi:noNamespaceSchemaLocation="https://apps.nextcloud.com/schema/apps/info.xsd">
+<info xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:noNamespaceSchemaLocation="https://apps.nextcloud.com/schema/apps/info.xsd">
     <id>photos</id>
     <name>Photos</name>
     <summary>Your memories under your control</summary>
     <description>Your memories under your control</description>
     <version>2.2.0</version>
     <licence>agpl</licence>
-	<author mail="skjnldsv@protonmail.com">John Molakvoæ</author>
+    <author mail="skjnldsv@protonmail.com">John Molakvoæ</author>
     <namespace>Photos</namespace>
     <category>multimedia</category>
     <types>
@@ -15,8 +15,8 @@
         <authentication />
     </types>
 
-	<website>https://github.com/nextcloud/photos</website>
-	<bugs>https://github.com/nextcloud/photos/issues</bugs>
+    <website>https://github.com/nextcloud/photos</website>
+    <bugs>https://github.com/nextcloud/photos/issues</bugs>
     <repository>https://github.com/nextcloud/photos.git</repository>
     <default_enable />
     <dependencies>
@@ -30,6 +30,11 @@
         </navigation>
     </navigations>
 
+    <commands>
+        <command>OCA\Photos\Command\UpdateReverseGeocodingFilesCommand</command>
+        <command>OCA\Photos\Command\MapMediaToLocationCommand</command>
+    </commands>
+
     <sabre>
         <collections>
             <collection>OCA\Photos\Sabre\RootCollection</collection>
@@ -39,4 +44,8 @@
             <plugin>OCA\Photos\Sabre\Album\PropFindPlugin</plugin>
         </plugins>
     </sabre>
+
+    <background-jobs>
+        <job>OCA\Photos\Jobs\AutomaticLocationMapperJob</job>
+    </background-jobs>
 </info>
--- a/lib/AppInfo/Application.php
+++ b/lib/AppInfo/Application.php
@@ -32,6 +32,7 @@ use OCA\Photos\Listener\NodeDeletedListener;
 use OCA\Photos\Listener\TagListener;
 use OCA\Photos\Listener\GroupUserRemovedListener;
 use OCA\Photos\Listener\GroupDeletedListener;
+use OCA\Photos\Listener\LocationManagerEventListener;
 use OCP\AppFramework\App;
 use OCP\AppFramework\Bootstrap\IBootContext;
 use OCP\AppFramework\Bootstrap\IBootstrap;
@@ -40,6 +41,9 @@ use OCP\Files\Events\Node\NodeDeletedEvent;
 use OCP\SystemTag\MapperEvent;
 use OCP\Group\Events\UserRemovedEvent;
 use OCP\Group\Events\GroupDeletedEvent;
+use OCP\Files\Events\Node\NodeWrittenEvent;
+
+require_once __DIR__ . '/../../vendor/autoload.php';
 
 class Application extends App implements IBootstrap {
 	public const APP_ID = 'photos';
@@ -77,6 +81,9 @@ class Application extends App implements IBootstrap {
 		$context->registerEventListener(UserRemovedEvent::class, GroupUserRemovedListener::class);
 
 		$context->registerEventListener(GroupDeletedEvent::class, GroupDeletedListener::class);
+
+		// Priority of -1 to be triggered after event listeners populating metadata.
+		$context->registerEventListener(NodeWrittenEvent::class, LocationManagerEventListener::class, -1);
 
 		$context->registerEventListener(SabrePluginAuthInitEvent::class, SabrePluginAuthInitListener::class);
 
--- a/lib/Command/MapMediaToLocationCommand.php
+++ b/lib/Command/MapMediaToLocationCommand.php
@@ -1,0 +1,116 @@
+<?php
+
+declare(strict_types=1);
+/**
+ * @copyright Copyright (c) 2022 Louis Chemineau <louis@chmn.me>
+ *
+ * @author Louis Chemineau <louis@chmn.me>
+ *
+ * @license AGPL-3.0-or-later
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+namespace OCA\Photos\Command;
+
+use OCP\IConfig;
+use OCP\IUserManager;
+use OCP\Files\IRootFolder;
+use OCP\Files\Folder;
+use OCA\Photos\Service\MediaLocationManager;
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Input\InputOption;
+use Symfony\Component\Console\Output\OutputInterface;
+
+class MapMediaToLocationCommand extends Command {
+	public function __construct(
+		private IRootFolder $rootFolder,
+		private MediaLocationManager $mediaLocationManager,
+		private IConfig $config,
+		private IUserManager $userManager,
+	) {
+		parent::__construct();
+	}
+
+	/**
+	 * Configure the command
+	 */
+	protected function configure(): void {
+		$this->setName('photos:map-media-to-location')
+			->setDescription('Reverse geocode media coordinates.')
+			->addOption('user', 'u', InputOption::VALUE_REQUIRED, 'Limit the mapping to a user.', null);
+	}
+
+	/**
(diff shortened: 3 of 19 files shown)


lib/AppInfo/Application.php
```diff
@@ -66,7 +67,9 @@ public function register(IRegistrationContext $context): void {
 		/** Register $principalBackend for the DAV collection */
 		$context->registerServiceAlias('principalBackend', Principal::class);
 		$context->registerEventListener(CacheEntryRemovedEvent::class, CacheEntryRemovedListener::class);
+		$context->registerEventListener(CacheEntryUpdatedEvent::class, LocationManagerNodeEventListener::class);
 		$context->registerEventListener(SabrePluginAuthInitEvent::class, SabrePluginAuthInitListener::class);
+		require_once __DIR__ . '/../../vendor/autoload.php';
```
artonge, line 72: Is this the correct way to load vendor files ?

lib/DB/PhotosFile.php
```diff
@@ -0,0 +1,91 @@
...
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+namespace OCA\Photos\DB;
+
+use OC\Metadata\FileMetadata;
+
+class PhotosFile {
```
artonge, line 30: Review hint: extracted from `AlbumFile` so it can be shared with `LocationFile`

appinfo/info.xml
```diff
@@ -1,22 +1,21 @@
 <?xml version="1.0"?>
-<info xmlns:xsi= "http://www.w3.org/2001/XMLSchema-instance"
-      xsi:noNamespaceSchemaLocation="https://apps.nextcloud.com/schema/apps/info.xsd">
+<info xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="https://apps.nextcloud.com/schema/apps/info.xsd">
     <id>photos</id>
     <name>Photos</name>
     <summary>Your memories under your control</summary>
     <description>Your memories under your control</description>
-    <version>2.1.0</version>
+    <version>2.2.0</version>
```
artonge, line 7: `2.3.0`

lib/Service/ReverseGeoCoderService.php
```diff
@@ -0,0 +1,167 @@
...
+		return $this->citiesMapping[$locationId] ?? '';
+	}
+
+	private function downloadCities1000(bool $force = false) {
+		if ($this->geoNameFolder->fileExists('cities1000.csv') && !$force) {
+			return;
+		}
+
+		// Download zip file to a tmp file.
+		$response = $this->clientService->newClient()->get("http://download.geonames.org/export/dump/cities1000.zip");
```
pulsejet, line 96: Just wanted to drop a note FYI: this database doesn't work very well if you query the closest data point, at least in urban US. This is mostly due to the large number of cities in metropolitan regions (I don't know what the case is in Europe / other countries). 

For instance, querying the closest location to me in the DB would either be Santa Monica or Westwood but I'm actually in LA. LA has 88 cities (and more unincorporated ones / villages in the DB), so photos from around the city would likely create a mess.

@artonge

artonge: Ah, I thought I answered you already.

What would you recommend ?

pulsejet: Hmm GitHub doesn't seem to notify correctly on review comments 🤔 

I've been trying to find a good solution to this for a long time now, but there's no easy one. There's no freely available dataset with cities and their boundaries that can be used directly. One idea is to use Overpass to dump OSM data into such a dataset (could use Geonames to get the list of cities to dump). Querying this efficiently might also be a bit challenging. I don't expect the dataset to be very large though; maybe a few gigabytes at most.

For now, I do think using Geonames is a good start, so I'll recommend going ahead with this 👍🏻 
Just wanted to drop this caution here in case someone can come up with a better idea.

artonge: :+1: thanks for sharing. Indeed, I don't see any other solution.

lib/Listener/LocationManagerNodeEventListener.php
```diff
@@ -0,0 +1,131 @@
...
+
+		if ($node instanceof File) {
+			if (!str_starts_with($node->getMimeType(), 'image')) {
+				return;
+			}
+
+			$callback($node->getId());
+		}
+	}
+}
```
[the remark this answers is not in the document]
artonge: Wrong statement, as we are dealing with shares here.

lib/Album/AlbumFile.php
```diff
@@ -23,19 +23,11 @@
...
+use OCA\Photos\DB\PhotosFile;
 
-class AlbumFile {
-	private int $fileId;
-	private string $name;
-	private string $mimeType;
-	private int $size;
-	private int $mtime;
-	private string $etag;
+class AlbumFile extends PhotosFile {
```
artonge, line 28: Review hint: extracted to `PhotosFile` so it can be shared with `LocationFile`

lib/Service/ReverseGeoCoderService.php
pulsejet: @artonge so I did finally get around to implementing this. The solution I settled on was to extract and simplify boundary data from the OSM planet database, then do an indexed db query on it. The results are _very_ accurate and the total data download size is at a reasonable ~80MB. Queries on this are surprisingly fast; on my slow-ish server it takes just ~1ms to query the entire planet. The downside is SQLite doesn't support geometric functions so it won't work there (but then who runs SQLite in production anyway).

I wrote a post about it if you're curious: https://varunpp2.medium.com/nextcloud-memories-putting-thousands-of-photos-on-the-world-map-1376cf4105a0

lib/DB/Location/LocationMapper.php
```diff
@@ -0,0 +1,119 @@
...
+			->from('mounts', 'mount')
+			->join('mount', 'filecache', 'file', $qb->expr()->eq('file.storage', 'mount.storage_id'))
+			->join('file', 'file_metadata', 'meta', $qb->expr()->eq('file.fileid', 'meta.id'))
+			->andWhere($qb->expr()->eq('mount.user_id', $qb->createNamedParameter($userId)))
+			// TODO: change 5 to the actual mimepart
+			->andWhere($qb->expr()->eq('file.mimepart', $qb->createNamedParameter($mimepart, IQueryBuilder::PARAM_INT)))
+			->andWhere($qb->expr()->eq('meta.group_name', $qb->createNamedParameter(self::METADATA_TYPE)))
+			->andWhere($qb->expr()->eq('meta.metadata', $qb->createNamedParameter($location)))
+			->executeQuery()
+			->fetchAll();
```
artonge, line 78: Translates to:

```sql
SELECT meta.metadata
FROM oc_mounts AS mount
	JOIN oc_filecache AS file ON file.storage = mount.storage_id
	JOIN oc_file_metadata AS meta ON file.fileid = meta.id
WHERE mount.user_id = '<a user ID>'
	AND file.mimempart = < the mime part for 'image' >
	AND meta.groupd_name = 'location';
```

lib/DB/Location/LocationMapper.php
```diff
@@ -0,0 +1,119 @@
...
+		$rows = $qb->selectDistinct('meta.metadata')
+			->from('mounts', 'mount')
+			->join('mount', 'filecache', 'file', $qb->expr()->eq('file.storage', 'mount.storage_id', IQueryBuilder::PARAM_INT))
+			->join('mount', 'file_metadata', 'meta', $qb->expr()->eq('file.fileid', 'meta.id', IQueryBuilder::PARAM_INT))
+			->andWhere($qb->expr()->eq('mount.user_id', $qb->createNamedParameter($userId)))
+			// TODO: change 5 to the actual mimepart
+			->andWhere($qb->expr()->eq('file.mimepart', $qb->createNamedParameter($mimepart, IQueryBuilder::PARAM_INT)))
+			->andWhere($qb->expr()->eq('meta.group_name', $qb->createNamedParameter(self::METADATA_TYPE)))
+			->executeQuery()
+			->fetchAll();
```
artonge, line 57: Translates to:

```sql
SELECT meta.metadata
FROM oc_mounts AS mount
	JOIN oc_filecache AS file ON file.storage = mount.storage_id
	JOIN oc_file_metadata AS meta ON file.fileid = meta.id
WHERE mount.user_id = '<a user ID>'
	AND file.mimempart = < the mime part for 'image' >
	AND meta.groupd_name = 'location';
```

lib/Service/ReverseGeoCoderService.php
```diff
@@ -0,0 +1,163 @@
...
+			file_put_contents($kdTreeTmpFileName, $kdTreeFileContent);
+			$fsTree = new FSKDTree($kdTreeTmpFileName, new ItemFactory());
+			$this->fsSearcher = new NearestSearch($fsTree);
+		}
+
+		$result = $this->fsSearcher->search(new Point([$latitude, $longitude]), 1);
+		return $this->getLocationNameForLocationId($result[0]->getId());
+	}
+
+	private function getLocationNameForLocationId(int $locationId): string {
```
artonge, line 71: Only loaded during the mapping, which is done in a background job, so it does not blow up on every request.

lib/Service/ReverseGeoCoderService.php
```diff
@@ -0,0 +1,167 @@
...
+			$this->buildKDTree();
+			$kdTreeFileContent = $this->geoNameFolder->getFile("cities1000.bin")->getContent();
+			$kdTreeTmpFileName = tempnam(sys_get_temp_dir(), "nextcloud_photos_");
+			file_put_contents($kdTreeTmpFileName, $kdTreeFileContent);
+			$fsTree = new FSKDTree($kdTreeTmpFileName, new ItemFactory());
+			$this->fsSearcher = new NearestSearch($fsTree);
+		}
+
+		$result = $this->fsSearcher->search(new Point([$latitude, $longitude]), 1);
+		return $result[0]->getId();
```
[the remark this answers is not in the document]
artonge: I made sure we can move to a more precise logic in the future. So resolving.

lib/Command/MapMediaToLocationCommand.php
```diff
@@ -0,0 +1,117 @@
...
+		$this->mediaLocationManager = $mediaLocationManager;
+		$this->userManager = $userManager;
+	}
+
+	/**
+	 * Configure the command
+	 *
+	 * @return void
+	 */
+	protected function configure() {
```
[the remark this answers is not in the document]
artonge: How to remove it later ?

lib/Jobs/AutomaticLocationMapperJob.php
```diff
@@ -0,0 +1,110 @@
...
+			if (time() - $startTime > 60 * 60) {
+				$this->config->setAppValue(Application::APP_ID, 'nextUserToScan', $user->getUID());
+				return;
+			}
+
+			$this->scanFilesForUser($user->getUID());
+		}
+
+		// TODO: kill this background job
+		$this->config->setAppValue(Application::APP_ID, 'nextUserToScan', '');
```
artonge, line 82: Any way to properly 'kill' this cronjob ?

lib/Jobs/AutomaticLocationMapperJob.php
```diff
@@ -0,0 +1,106 @@
...
+				continue;
+			}
+
+			// Stop if execution time is more than one hour.
+			if (time() - $startTime > 60 * 60) {
+				return;
+			}
+
+			$this->scanFilesForUser($user->getUID());
+			$this->config->setAppValue(Application::APP_ID, 'lastMappedUser', $user->getUID());
```
artonge, line 77: @ArtificialOwl any way to properly kill this job ? Currently, after having loop over every user, it will still get called regularly, but won't do anything.

ArtificialOwl (approved):

Issue #1292, opened 2022-09-24
Internal server error at cloud.nextcloud com


rakekniven:
**Describe the bug**

Selecting photos app shows internal server error.

**To Reproduce**
1. I opened https://cloud.nextcloud.com/
2. Logged in
3. Selected photos app
4. Internal server error is shown

**Expected behavior**
Photos app should open.

**Screenshots**
<img width="807" alt="Bildschirmfoto 2022-09-24 um 17 43 43" src="https://user-images.githubusercontent.com/2069590/192106834-27c41bf9-e9bc-484b-a8fc-36c1f84c1d2e.png">

**Desktop (please complete the following information):**
 - OS: macOS
 - Browser: firefox
 - Version: 104

NC version: 25.0.0 RC 1

**Additional context**
Other users at community chat confirmed error.

skjnldsv:
Hi, Which section?

nickvergessen:
Log entry with above request_id `subz8Hzjsm8u9RaBLlKY`:
```json
{
  "reqId": "subz8Hzjsm8u9RaBLlKY",
  "level": 3,
  "time": "2022-09-24T15:43:32+00:00",
  "remoteAddr": "…",
  "user": "…(guest app account)",
  "app": "index",
  "method": "GET",
  "url": "/apps/photos/",
  "message": "No create permission for folder",
  "userAgent": "…",
  "version": "25.0.0.13",
  "exception": {
    "Exception": "OCP\\Files\\NotPermittedException",
    "Message": "No create permission for folder",
    "Code": 0,
    "Trace": [
      {
        "function": "newFolder",
        "class": "OC\\Files\\Node\\Folder",
        "type": "->",
        "args": [
          "/Photos"
        ]
      },
      {
        "file": "/var/www/cloud.nextcloud.com/nextcloud/lib/private/Files/Node/LazyFolder.php",
        "line": 72,
        "function": "call_user_func_array",
        "args": [
          [
            {
              "__class__": "OC\\Files\\Node\\Folder"
            },
            "newFolder"
          ],
          [
            "/Photos"
          ]
        ]
      },
      {
        "file": "/var/www/cloud.nextcloud.com/nextcloud/lib/private/Files/Node/LazyFolder.php",
        "line": 419,
        "function": "__call",
        "class": "OC\\Files\\Node\\LazyFolder",
        "type": "->",
        "args": [
          "newFolder",
          [
            "/Photos"
          ]
        ]
      },
      {
[...]

PR #1325, opened 2022-10-11
Make sure the default photos folder exists if none

skjnldsv:
Fix https://github.com/nextcloud/photos/issues/1292


artonge (approved):
Not sure why it is different, but if it works, then ok :)

skjnldsv:
> Not sure why it is different, but if it works better, then ok :)

We need to create the default folder somewhere :grin:

PR #1334, opened 2022-10-13
Add location grouping views

artonge:
Needs: https://github.com/nextcloud/photos/pull/1335

Screenshot 2022-10-26 at 15-22-29 Photos - Nextcloud

Screenshot 2022-10-26 at 15-30-48 Photos - Nextcloud


artonge:
@jancborchardt for early feedback :)

artonge:
Or @nimishavijay

nimishavijay:
Looks great! :) Only one point: "Locations" --> Places 

Is there some sort of map view planned/possible?

artonge:
> Looks great! :) Only one point: "Locations" --> Places
> 
> Is there some sort of map view planned/possible?

Already done by the Map app

nimishavijay:
> Already done by the Map app

It would be nice to have that view in Photos as well, since it may not be super discoverable from Maps :) possibly as an enhancement for the future?

artonge:
> It would be nice to have that view in Photos as well, since it may not be super discoverable from Maps :) possibly as an enhancement for the future?

IIRC it was discussed and considered as not in the scope of the photos app. I don't have a strong opinion, but having it in the map app seems enough to me.

cypress/e2e/albums.cy.js
```diff
@@ -128,21 +128,21 @@ describe('Manage albums', () => {
...
+		cy.get('form [name="place"]').clear().type('New place')
 		cy.contains('Save').click()
 
-		cy.contains('New location')
+		cy.contains('New place')
 
 		cy.reload()
 
-		cy.contains('New location')
+		cy.contains('New place')
```
skjnldsv, line 141: It seems a bit too broad. We can merge but I would make sure to check against the specific section that contains this string, not the entire page

skjnldsv:
Followups:
- [ ] The location column of the photos_albums table is still present, since this is now a dav property, they should be migrated  image
image
- [ ] Can you clarify if the albums should be displayed in the `Places` section? Because I find the two confusing. Files can have a location metadata, but Albums can have a location property. Only the files metadata  seems to be used.
- [ ] You manually have to run `photos:map-media-to-place` for the files location to be updated, right? Is there no on-upload or on-update file processing?

skjnldsv (approved):

cypress/e2e/albums.cy.js
artonge: Was indeed too broad :)
https://github.com/nextcloud/photos/pull/1657

artonge:
>     * [ ]  The location column of the photos_albums table is still present, since this is now a dav property, they should be migrated

Not sure. See answer for the second followup.

>     * [ ]  Can you clarify if the albums should be displayed in the `Places` section? Because I find the two confusing. Files can have a location metadata, but Albums can have a location property. Only the files metadata  seems to be used.

'Place' for a file !== 'Location' for an album

Each file can have a place from its GPS coordinate.
Each album can have a location set by the user.

>     * [ ]  You manually have to run `photos:map-media-to-place` for the files location to be updated, right? Is there no on-upload or on-update file processing?

A background job was added to do the mapping in the background after the migration to the new version. The command is here to make it faster or to remap a user's files.

artonge:
Thanks for taking care of merging :)

PR #1337, opened 2022-10-13
Listen to more events for albums

artonge:
Add handling for `UserDeletedEvent` and `ShareDeletedEvent`

Any other event we should listen too ?


lib/Listener/AlbumsManagementFileEventListener.php
```diff
@@ -0,0 +1,64 @@
...
+			// Delete all user's albums.
+			$albums = $this->albumMapper->getForUser($event->getUser()->getUID());
+			foreach ($albums as $album) {
+				$this->albumMapper->delete($album->getId());
+			}
+		}
+
+		if ($event instanceof ShareDeletedEvent) {
+			$receiverId = $event->getShare()->getSharedWith();
+			$this->forEachSubNode(
```
artonge, line 41: Going through each subnodes of the share to remove them from albums.
Could be problematic performance wise, no ? Would it make sense to put that into a job ?

lib/Album/AlbumMapper.php
```diff
@@ -273,6 +273,32 @@ public function removeFile(int $albumId, int $fileId): void {
...
+			->where($query->expr()->eq("owner_id", $query->createNamedParameter($ownerId)))
+			->andWhere($query->expr()->eq("file_id", $query->createNamedParameter($fileId, IQueryBuilder::PARAM_INT)));
+		$query->executeStatement();
+
+		// Remove any occurrence of fileId when owner is ownerId.
+		$query = $this->connection->getQueryBuilder();
+		$query->delete("photos_albums_files")
+			->where($query->expr()->eq("owner_id", $query->createNamedParameter($ownerId)))
+			->andWhere($query->expr()->eq("file_id", $query->createNamedParameter($fileId, IQueryBuilder::PARAM_INT)));
+		$query->executeStatement();
```
artonge, line 290: Is there a way to get some data from a delete call ? Ideally I would get album ids from the delete call, but not sure if that is possible.

lib/Listener/AlbumsManagementFileEventListener.php
artonge: Discussed, and an event will be emitted for each deleted entries by the server.

simonspa:
@artonge I think this is useful and will make the response to system changes more complete. However, I think this needs a rebase because of the additional events added in #1601 and the change to the cache listener event in #1598

lib/Listener/AlbumsManagementEventListener.php
```diff
@@ -0,0 +1,64 @@
...
+
+class AlbumsManagementEventListener implements IEventListener {
+	private AlbumMapper $albumMapper;
+
+	public function __construct(AlbumMapper $albumMapper) {
+		$this->albumMapper = $albumMapper;
+	}
+
+	public function handle(Event $event): void {
+		if ($event instanceof CacheEntryRemovedEvent) {
```
simonspa, line 23: This is exactly the same as https://github.com/nextcloud/photos/blob/master/lib/Listener/NodeDeletedListener.php, so we should probably just listen to this event also there?

I had this in but you requested to remove it in #1598 IIRC

skjnldsv (approved):

lib/Listener/AlbumsManagementEventListener.php
```diff
@@ -0,0 +1,111 @@
...
+	}
+
+	private function forEachSubNode(Node $node, callable $callback): void {
+		if ($node instanceof Folder) {
+			foreach ($node->getDirectoryListing() as $subNode) {
+				$this->forEachSubNode($subNode, $callback);
+			}
+		}
+
+		if ($node instanceof File) {
```
come-nc, line 103:
```suggestion
		} elseif ($node instanceof File) {
```

lib/Listener/AlbumsManagementEventListener.php
```diff
@@ -0,0 +1,111 @@
...
+			// Get all group-shared albums that are not directly shared with the removed user in addition
+			$albums = array_udiff($albums_group, $albums_user, fn ($a, $b) => ($a->getId() - $b->getId()));
+
+			// Remove their photos from theses albums:
+			foreach ($albums as $album) {
+				$this->albumMapper->removeFilesForUser($album->getId(), $event->getUser()->getUID());
+			}
+		}
+
+		if ($event instanceof GroupDeletedEvent) {
```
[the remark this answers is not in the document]
artonge: The other kind of events? They are above

PR #1340, opened 2022-10-14
This is Photos 2.0

skjnldsv:
Since we advertise photos 2, we need to make sure it is.

WIll bump minor after this is merged for 26


szaimen (approved):

skjnldsv:
https://github.com/nextcloud/photos/pull/1342

PR #1341, opened 2022-10-14
Fix nomedia exclusion

### LINKED ISSUE #1338: .nomedia and .noimage have no effect in Photos and All Media
**Describe the bug**
I have created .noimage and .nomedia files in several folders that I wish to have excluded from Photos. The images within those folders still appear within Photos. As far as I understand from reading the other .nomedia issues here, this should work.

**To Reproduce**
Create a .nomedia or .noimage file in a folder.

**Expected behavior**
Images in a folder or subfolder with a .nomedia file should not appear in Photos.

**Additional context**
Is there a way to force a re-scan? I tried occ files:scan --all but this did not work, and I could not find any reference to .nomedia in the Nextcloud documentation. 

Server version is 25.0.0 RC3

skjnldsv:
@artonge I found some logic issues in the code.

I think we should do a round of checks to make sure the files we locally store in components (like [here](https://github.com/nextcloud/photos/blob/912dc1976e281f52d49ca038cca98c78c1873f33/src/mixins/FetchFilesMixin.js#L92-L96)), are always vefified against the store for undefined. We might have fileIds that are NOT allowed to be displayed like here :)

Basically doing checks on that kind of stuff 
https://github.com/nextcloud/photos/blob/a48b3a1fc7a9b2b72740a829a479f4ca6512b0e8/src/mixins/FilesByMonthMixin.js#L32-L38

fix #1338


marcelklehr (approved):

skjnldsv:
We should yep

Issue #1368, opened 2022-10-16
Photo is triggering bruteforce protection


CarlSchwan:
**Describe the bug**

**To Reproduce**
1. Enable desktop client
2. Try to login
3. Get blocked 20 seconds later
4. Clear bruteforce attempts on the server
5. Get blocked again 20 seconds later

**Expected behavior**
No bruteforce protection triggered for normal usage

```
[...]

PR #1370, opened 2022-10-17
Limit auth backend to public albums

artonge:
Fix https://github.com/nextcloud/photos/issues/1368

And clean `PublicAlbumAuthBackend.php` a bit.

### DIFF
--- a/lib/Listener/SabrePluginAuthInitListener.php
+++ b/lib/Listener/SabrePluginAuthInitListener.php
@@ -43,6 +43,11 @@ class SabrePluginAuthInitListener implements IEventListener {
 		}
 
 		$server = $event->getServer();
+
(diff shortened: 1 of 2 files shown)


lib/Listener/SabrePluginAuthInitListener.php
```diff
@@ -43,6 +43,11 @@ public function handle(Event $event): void {
 		}
 
 		$server = $event->getServer();
+
+		if (!str_starts_with($server->getRequestUri(), 'photospublic/')) {
+			return;
+		}
+
```
artonge, line 50: I could not reproduce as my desktop client do not show the main dialog to add an account right now :thinking: 
@CarlSchwan could you confirm that this fixes the issue ?

skjnldsv (approved):

PR #1372, opened 2022-10-18
Prevent building prefixPath on loading

artonge:
Follow up of https://github.com/nextcloud/photos/pull/1341 as `currentUser` is null on public pages

Supersede https://github.com/nextcloud/photos/pull/1371


skjnldsv (approved):
:ship:

PR #1388, opened 2022-10-19
Fix fetching album content when navigating

artonge:
When navigating between albums:

- The call to `newFileMenuEntry` was causing an error, because the entry were already existing as we did not remove it when destroying `AlbumContent`
- `fetchAlbumContent` was never called as the album's `fileName` did not change
- `addFilesToAlbum` was wrongly incrementing the items count, so I created another mutation called `setAlbumFiles` that only set the album's files.


src/store/albums.js
```diff
@@ -77,6 +77,29 @@ const mutations = {
...
+	 * @param {string} data.albumName the album id
+	 * @param {string[]} data.fileIds list of files
+	 */
+	setAlbumFiles(state, { albumName, fileIds }) {
+		const albumFiles = state.albumsFiles[albumName] || []
+		state.albumsFiles = {
+			...state.albumsFiles,
+			[albumName]: [
+				...albumFiles,
+				...fileIds.filter(fileId => !albumFiles.includes(fileId)), // Filter to prevent duplicate fileId.
```
skjnldsv, line 94: We should really prefer using Set for performances reasons :)
https://jsben.ch/JNnpx

skjnldsv: image

artonge: Can we address that later during a refactoring ?

artonge: But you made your point :)

artonge: I actually already made the change in `collectionStoreFactory` that I want to use for albums and shareAlbums

https://github.com/nextcloud/photos/blob/master/src/store/collectionStoreFactory.js#L76-L83

artonge: But truth for truth, I am actually thinking of more generic store structure instead of having dedicated stores with `collectionStoreFactory`.

artonge: But first, make it work, then optimize

skjnldsv: > Can we address that later during a refactoring ?

Absolutely :tada:

skjnldsv: I'm wondering if there isn't some tools out there that can detect such things :thinking: 
That seems not too crazy. Easy recommendations tbh :see_no_evil:

skjnldsv (approved):

PR #1390, opened 2022-10-19
Downgrade to @nc/vue v7.0.0

artonge:
Focus traps cause an infinite recursion, which make opening any modal laggy, this also cause e2e tests to fail.

Downgrade until a proper solution is found.

https://github.com/nextcloud/nextcloud-focus-trap/issues/2


skjnldsv:
Hum, that seems like the other way around ?

artonge:
What do you mean ? That v7.0.1 should contain a fix ? Not in my testing.

skjnldsv:
> What do you mean ? That v7.0.1 should contain a fix ? Not in my testing.

I mean, which modal are we opening?

artonge:
When creating an album or adding collaborators

artonge:
Approve then ? :D

skjnldsv (approved):
😁😁

PR #1417, opened 2022-10-24
Fix for ".nomedia" paths and deletes in Viewer

jkellerer:
This PR is a collection of ~~3~~ 2 fixes in 2 commits:
* ~~Source URL could be invalid and this broke sidebar, download and file-editing in the Viewer. The URL derives from `filename` but this was not consistently built (sometimes with, sometimes without prefix). This change ensures `filename` is consistent and a valid URL can always be built~~ - moved to #1419.
* `$store.appendFiles` may filter files in `.nomedia` paths but the callers of `appendFiles` didn't evaluate whether the method really appended all files, leading to broken references. This change ensures that only files that were really appended will be considered.
* A `Delete File` action inside the `Viewer` was not propagated to the `Photos` app (the view still contained deleted files and caused failures when trying to get the file from the server).

Hope this found useful. Had only fixed the most serious problems I found when trying the app. 
Feel free to reject or accept the PR or individual commits of it.

### DIFF
--- a/src/store/files.js
+++ b/src/store/files.js
@@ -150,9 +150,11 @@ const actions = {
 	 *
 	 * @param {object} context the store mutations
 	 * @param {Array} files list of files
+	 * @return {Array} the appended files
 	 */
 	appendFiles(context, files = []) {
 		context.commit('updateFiles', files)
+		return files.filter(file => !!state.files[file.fileid])
 	},
 
 	/**
--- a/src/store/files.js
+++ b/src/store/files.js
@@ -148,12 +148,13 @@ const actions = {
 	/**
 	 * Append or update given files
 	 *
-	 * @param {object} context the store mutations
+	 * @param {object.commit} commit the store mutation commit function
+	 * @param {object.state} state the store state
 	 * @param {Array} files list of files
 	 * @return {Array} the appended files
 	 */
-	appendFiles(context, files = []) {
-		context.commit('updateFiles', files)
+	appendFiles({ commit, state }, files = []) {
+		commit('updateFiles', files)
 		return files.filter(file => !!state.files[file.fileid])
 	},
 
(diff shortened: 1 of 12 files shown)


skjnldsv (changes requested):
Hey, the [`Fix 'source' URL in viewer`](https://github.com/nextcloud/photos/pull/1417/commits/e787d7d5008e41fe681d44ddab414faf525fa0ab) part seems far too complicated.
I have plans to fix this for 26. Until then, I would like to keep the code simple.

Please strip this out from this PR and open a dedicated one for it so we can discuss this separately :)

skjnldsv (approved):

src/store/files.js
```diff
@@ -150,9 +150,11 @@ const actions = {
 	 *
 	 * @param {object} context the store mutations
 	 * @param {Array} files list of files
+	 * @return {Array} the appended files
 	 */
 	appendFiles(context, files = []) {
 		context.commit('updateFiles', files)
+		return files.filter(file => !!state.files[file.fileid])
```
skjnldsv, line 157: `state` is undefined

skjnldsv, line 157: The proper way to do it would be

```suggestion
	 * @return {Array} the appended files
	 */
	appendFiles({ commit, state }, files = []) {
		commit('updateFiles', files)
		return files.filter(file => !!state.files[file.fileid])
```

jkellerer: or `context.state.files`? .. had taken it from line 181, where `state` is also used directly.

jkellerer: I see your proposal.. context has commit and files.. yes that makes sense

skjnldsv: > or `context.state.files`? .. had taken it from line 181, where `state` is also used directly.

Care to also change that? :see_no_evil: 
But yes, `context.state` also works :)

simonspa:
@artonge anything missing for this pr? Looks complete and has an approval.

If Cypress fail is related is unclear since the logs expired

PR #1419, opened 2022-10-25
Fix 'source' URL in viewer

jkellerer:
This PR concentrates on `filename` being consistent and includes app and user so that source URL can be built properly. This allows to use the full viewer (with fileInfo as input) at more places.

Prior to this change, source URL could be invalid and this broke sidebar, download and file-editing in the Viewer. The URL derives from `filename` but this was not consistently built (sometimes with, sometimes without prefix).

Since the viewer may also need different paths, all input to the viewer is mapped so that it can be adjusted as needed (at a single place).

Note: In my tests albums views still had an issue (e.g. forcing the use of the URL with `hasPreview: false` fails), but this may be related to https://github.com/nextcloud/viewer/issues/1282

### DIFF
--- a/src/components/Collection/CollectionContent.vue
+++ b/src/components/Collection/CollectionContent.vue
@@ -68,6 +68,7 @@ import FilesListViewer from '.././FilesListViewer.vue'
 import File from '.././File.vue'
 import FolderIllustration from '../../assets/Illustrations/folder.svg'
 import SemaphoreWithPriority from '../../utils/semaphoreWithPriority.js'
+import { toViewerFileInfo } from '../../utils/fileUtils.js'
 
 export default {
 	name: 'CollectionContent',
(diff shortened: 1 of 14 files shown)


skjnldsv (changes requested):
Blocking, I'm investigating a different approach from Viewer for 26, we'll see if that is really necessary to have that heavy changes to fix things :)

vprogramer:
Hello. I need help. I don't see previews and images, but I can download photos. I deployed nextcloud and copied data and database from another old nextcloud. After the installation was completed, everything worked fine, but the photos did not load. What to do?
I've tried deleting these photos and uploading them again, I've tried creating new previews, but it didn't work. But if I take a screenshot from a photo, then I can upload it.
image

skjnldsv:
@vprogramer please open your own issue and fill the reporting template! :wink:

PR #1491, opened 2022-11-23
Prevent throwing an error in node deletion hook

artonge:
Not satisfied by this catch all approach. Maybe there are better ways to do that ? 

Fix https://github.com/nextcloud/server/issues/35287


PVince81 (approved):
:+1: ok for now

for ideas see https://github.com/nextcloud/server/issues/35287#issuecomment-1324938659

Issue #1137, opened 2022-07-17
[BITV/A11Y] Title not updating when changing sections in navigation


AndyScherzinger:
Steps to reproduce:

Open Photos app (also on c.nc.com)
Click on the "Videos" section on the left
Expected result: document title also says "Videos - Nextcloud" or "Videos - Photos - Nextcloud".

Actual result: document title always stays on"Photos - Nextcloud"

PVince81:
@skjnldsv can you maybe check this one ?

JuliaKirschenheuter:
@AndyScherzinger , @PVince81 Could you please post some screenshots / video captures?

PVince81:
@JuliaKirschenheuter steps to reproduce:

1. Open Photos app (also on c.nc.com)
2. Click on the "Videos" section on the left

Expected result: document title also says "Videos - Nextcloud" or "Videos - Photos - Nextcloud".

Actual result: document title always stays on"Photos - Nextcloud"

PR #1510, opened 2022-11-28
Update title when changing sections in navigation

JuliaKirschenheuter:
Fixes https://github.com/nextcloud/photos/issues/1137

Reviewers: please make a double-check


ChristophWurst:
> Reviewers: please make a double-check

Of what? Are there known issues?

JuliaKirschenheuter:
> > Reviewers: please make a double-check
> 
> Of what? Are there known issues?

I'm not aware with some places like https://github.com/nextcloud/photos/blob/cf9e0a30832a7a7559db655b4ee67b20121093c7/src/router/index.js#L166-L169

please click everywhere to be sure that is right

src/router/index.js
```diff
@@ -105,11 +127,21 @@ const router = new Router({
 			props: route => ({
 				albumName: route.params.albumName,
 			}),
+			meta: {
+				rootTitle: (to) => {
+					return `${t('photos', 'Album')} ${to.params.albumName}`
```
ChristophWurst, line 132: You are concatenating a translated string. Always use placeholders inside the translation string instead.

```suggestion
					return t('photos', 'Album {title}', { title: to.params.albumName })
```

ChristophWurst (changes requested):

src/router/index.js
```diff
@@ -74,6 +76,11 @@ const router = new Router({
 			props: route => ({
 				rootTitle: t('photos', 'All your media'),
 			}),
+			meta: {
```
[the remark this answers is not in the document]
JuliaKirschenheuter: Unfortunately that doesn't works this way. ``` props ``` are transmitted from here to component and won't be saved in the route. Because of this i need another solution which will be saved in a route.

ChristophWurst (approved):

src/router/index.js
```diff
@@ -215,8 +287,23 @@ const router = new Router({
 				rootTitle: route.params.faceName,
 				faceName: route.params.faceName,
 			}),
+			meta: {
+				rootTitle: (to) => {
+					return t('photos', 'Face {title}', { title: to.params.rootTitle })
```
skjnldsv, line 292: Not convinced of that wording, what would be `rootTitle`, a name?

Then it would return `Face John` ? Seems weird :see_no_evil:

ChristophWurst: ``{title}'s face``

JuliaKirschenheuter: @ChristophWurst should i change it to ``` return t('photos', "{title}'s face", { title: to.params.rootTitle }) ```
but then there is only EN version. Or just "Faces"

skjnldsv (approved):

src/router/index.js
ChristophWurst: Why is there only an English version if you run the string through translation?

JuliaKirschenheuter: i mean would a " 's " translated correctly? If yes - i will do it this way

ChristophWurst: Yes. Translators can change that part of the string, e.g. to ``{title}s Gesicht`` for German.

Issue #1596, opened 2023-01-21
[Bug] Sharing Albums with Groups Doesn't Work


iamjameswalters:
**Describe the bug**
When you share an album with a group of users, those users cannot add photos to the album. My server's been crashing when trying to do this.

**To Reproduce**
1. Create a user group
2. Create a photo album
3. Share the album with the user group
4. Try to add a photo to the shared album as a member of the user group that is not the album owner

**Expected behavior**
Members of user groups with shared permissions should be able to add photos.

**Desktop (please complete the following information):**
 - OS: Ubuntu 22.04
 - Browser: Any (tested with Firefox and Brave)
 - Version: n/a

**Smartphone (please complete the following information):**
- n/a

**Additional context**
When sharing with an individual user instead of a user group, this works as expected.

simonspa:
For reference:

```
`Undefined property: OCA\Photos\Sabre\Album\SharedAlbumRoot::$userFolder at /var/www/html/apps-extra/photos/lib/Sabre/Album/AlbumRoot.php#98`
```

simonspa:
The problem appears to be that [here](https://github.com/nextcloud/photos/blob/master/lib/Sabre/Album/SharedAlbumRoot.php#L54-L56) it is only checked if the *user ID* is among the collaborators, but there is no check if the current user is member of a *group* that is among the collaborators:

```php
		$collaboratorIds = array_map(
			fn ($collaborator) => $collaborator['type'].':'.$collaborator['id'],
			$this->albumMapper->getCollaborators($this->album->getAlbum()->getId()),
		);
		if (!in_array(AlbumMapper::TYPE_USER.':'.$this->userId, $collaboratorIds)) {
			return false;
		}
```

Unfortunately I don't know how to solve this since in `SharedAlbumRoot` I don't have access to a `IGroupManager` to check if the user is a member. Alternatively the `AlbumMapper` could return a list of all users, directly resolving the groups - however, this is not how the `getCollaborators(...)` method was designed and might have other problems. Also, there is also public links which need to fit in.

iamjameswalters:
Thanks for working on this @simonspa !

PR #1597, opened 2023-01-21
Remove photos from collaborators once they are removed

simonspa:
Right now photos stay in the album even after collaborators who added them either remove themselves from the album or are removed by the album owner.

Apart from the fact that accessing any photo that is not your own right now is utterly broken, I think when being removed from an album, also the respective user photos should be removed - they still own this data.

This MR attempts to implement this.

Please not that group sharing and photo adding is
a) not covered by this yet and
b) is completely broken (see https://github.com/nextcloud/photos/issues/1596)


artonge (approved):

artonge:
Rebasing to trigger CI

PR #1598, opened 2023-01-22
Listen to NodeDeletedEvent to remove photos from albums

### LINKED ISSUE #1535: Album photo count incorrect
**Describe the bug**
When you delete a photo from an album that was added with "Add photos to this album", and then delete the photo from Files, Photos 2.0 does not correct the counter with the deleted photo. You also can't dereference the deleted file so the number is permanently incorrect.

**To Reproduce**
1. Create a new album
2. Add a photo that already exists on Nextcloud
3. Go to files and delete the photo
4. Go back to albums. The thumbnail for the deleted photo is still present, and the counter shows 1 photo, but it should be 0 since the only photo added to the album was deleted

**Expected behavior**
When deleting a photo in Files that is in an album, the photo should either be kept in the album and still accessible, or deleted entirely from the album

simonspa:
This MR adds a listener for `NodeDeletedEvent` which is emitted when deleting (or moving to trash bin) a file from *Files* or *Photos* itself. I don;t understand when the current `CacheEntryRemoved` event is emitted so I left it in - but someone with more insight into the NC ecosystem should probably have a look at this.

Also, I noted that this event `NodeDeletedEvent` is not documented in https://docs.nextcloud.com/server/latest/developer_manual/basics/events.html and the documented events such as `NodeRemovedFromCache` do not seem to be relevant here.

This fixes #1535

### DIFF
--- a/lib/AppInfo/Application.php
+++ b/lib/AppInfo/Application.php
@@ -34,6 +34,7 @@ use OCP\AppFramework\Bootstrap\IBootContext;
 use OCP\AppFramework\Bootstrap\IBootstrap;
 use OCP\AppFramework\Bootstrap\IRegistrationContext;
 use OCP\Files\Cache\CacheEntryRemovedEvent;
+use OCP\Files\Events\Node\NodeDeletedEvent;
 
 class Application extends App implements IBootstrap {
 	public const APP_ID = 'photos';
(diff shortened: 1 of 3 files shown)


artonge:
Will check why `CacheEntryRemoved` is not enough

simonspa:
Somehow for me this was never triggered - we can possibly remove it completely then>?

artonge:
I remember now. `CacheEntryRemoved` is not triggered because the file is "only" moved to the trashbin and not fully deleted. At the time, my opinion was that it should not remove the file from the album if it was only in the trash. But I am not sure that I agree with me anymore. Let's go with `NodeDeletedEvent`. Can you remove reference to `CacheEntryRemoved` ?

simonspa:
Done so. :)

artonge (approved):
:ok_hand:

artonge:
Unsure about that one. What are you trying to achieve ?

PR #1599, opened 2023-01-22
Fix upload of pictures into album

simonspa:
Fixing two issues related to uploading pictures to albums:

* Album IDs were compared as strings but are `int`s
* Attempted to fetch user folder for storing uploaded photos from `$this` but it's only a local variable

There are still more issues to be solved here:
* Uploaded pictures do not show up in preview directly (i.e. need to reload page?)
* Still issues in adding pictures to group-shared albums (even when not uploading file we end in `createFile`...)

### DIFF
--- a/lib/Controller/PreviewController.php
+++ b/lib/Controller/PreviewController.php
@@ -101,7 +101,7 @@ class PreviewController extends Controller {
 
 		if (\count($nodes) === 0) {
 			$receivedAlbums = $this->albumMapper->getAlbumsForCollaboratorIdAndFileId($user->getUID(), AlbumMapper::TYPE_USER, $fileId);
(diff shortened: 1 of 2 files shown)


artonge (approved):
:+1:

simonspa:
@artonge nextcloud bot doesn't like me - but I think this should be backported, just as the other coming pulls.

artonge:
> @artonge nextcloud bot doesn't like me - but I think this should be backported, just as the other coming pulls.

Try again with `/backport to ...` ;)

simonspa:
Yihaaa, they like me now! :tada: Thanks :smile:

artonge:
Not so much apparently ^^
This needs to be done manually because of conflicts.

PR #1600, opened 2023-01-22
Fix sharing with group

### LINKED ISSUE #1596: [Bug] Sharing Albums with Groups Doesn't Work

simonspa:
This MR fixes adding photos to an album shared to a group of users. The essential problem was that only collaboration via `userId` was checked.

This adds a new method to `AlbumMapper` to check for any given user ID of the user is a collaborator, no matter whether this is via direct sharing or group sharing.

This fixes #1596

### DIFF
--- a/lib/Album/AlbumMapper.php
+++ b/lib/Album/AlbumMapper.php
@@ -349,7 +349,7 @@ class AlbumMapper {
 		foreach ($rows as $row) {
 			switch ($row['collaborator_type']) {
 				case self::TYPE_USER:
-					if (!strcmp($row['collaborator_id'], $userId)) {
+					if ($row['collaborator_id'] === $userId) {
 						return true;
 					}
 					break;
--- a/lib/Album/AlbumMapper.php
+++ b/lib/Album/AlbumMapper.php
@@ -349,7 +349,7 @@ class AlbumMapper {
 		foreach ($rows as $row) {
 			switch ($row['collaborator_type']) {
 				case self::TYPE_USER:
-					if (strcmp($row['collaborator_id'], $userId)) {
+					if (!strcmp($row['collaborator_id'], $userId)) {
 						return true;
 					}
 					break;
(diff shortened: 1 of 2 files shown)


lib/Album/AlbumMapper.php
```diff
@@ -332,6 +332,40 @@ public function getCollaborators(int $albumId): array {
...
+		$query->select("collaborator_id", "collaborator_type")
+			->from("photos_albums_collabs")
+			->where($query->expr()->eq('album_id', $query->createNamedParameter($albumId, IQueryBuilder::PARAM_INT)));
+
+		$rows = $query->executeQuery()->fetchAll();
+
+		foreach ($rows as $row) {
+			switch ($row['collaborator_type']) {
+				case self::TYPE_USER:
+					if (!strcmp($row['collaborator_id'], $userId)) {
```
artonge, line 352: Why not use `===` instead of `strcmp` ?

simonspa: No particular reason, I've seen `strcmp` being used elsewhere in similar situations. Shall I change it?

artonge: Yes, as it is more readable with `===`.
I only used `strcmp` for comparison during array sorting :).

simonspa: Done :)

artonge (approved):
:rocket:

PR #1601, opened 2023-01-22
Correctly listen to group change events

simonspa:
In order for group sharing to properly work, we also need to listen to changes in the group:

* A user is removed from the group (`Group\UserRemovedEvent`): We need to remove all pictures from this user from all albums *unless they are shared also directly with the user*
* A group is dropped completely (`Group\GroupDeletedEvent`): We need to remove the group as collaborator from all albums *and* remove all files added by users in this group *unless they also have a direct share of the album*

This MR builds on top of my other MRs and I'd be happy to rebase this branch once the others are merged.


artonge:
> * A user is removed from the group (`Group\UserRemovedEvent`): We need to remove all pictures from this user from all albums _unless they are shared also directly with the user_
> 
>     * A group is dropped completely (`Group\GroupDeletedEvent`): We need to remove the group as collaborator from all albums _and_ remove all files added by users in this group _unless they also have a direct share of the album_

I think that we should delete the photo even if the album owner has access to the photo through another mean. It makes things simpler and is more

simonspa:
> I think that we should delete the photo even if the album owner has access to the photo through another mean. It makes things simpler and is probably what a user would expect.

I would expect the opposite actually. Imagine:

I create an album "NC Photos Hackweek" and share it with you. You add your pictures to it. Then we decide to also share it with the rest of the NC Photos team as group share. Later we realize that that's not a good idea and remove the group share again. I would expect your pictures to stay in because you are still a shareholder (sorry, collaborator) of the album. Your pictures would have been removed *only* because you *also* happened to be in the NC Photos team.

simonspa:
...to add: the issue is that we cannot determine by which means (group or direct share) a photo has been added. (and it's good because knowing this would make things even more complicated... :smile: )

artonge:
I misunderstood your initial comment, I thought *"unless they are shared also directly with the user"* was referring to a classic file share.

Other PR seems bundled inside this one. I'll review the code of this one once the other are merged :).

simonspa:
Ah, sorry, no - I stayed completely within the Photos realm. :) I agree that folding in other sharing types into this is neither sensible nor easily feasible.

I'll rebase this once the others are merged.

lib/Listener/GroupDeletedListener.php
```diff
@@ -0,0 +1,48 @@
...
+			return;
+		}
+        
+        // Get all shared albums for this group:
+        $albums_group = $this->albumMapper->getSharedAlbumsForCollaboratorWithFiles($event->getGroup()->getGID(), AlbumMapper::TYPE_GROUP);
+
+        // Get all users of this group:
+        $users = $event->getGroup()->getUsers();
+
+        foreach ($users as $user) {
```
artonge, line 28: The algo seem convoluted.

- You have the list of albums for a group
- You have the list of users for a group
- For each albums:
    - Remove the group from the collaborators list
    - For each user
         - If the use is not in the collaborator list, remove its files.

I don't think you need to query the DB more than once.

Let me know if I am missing something.

lib/Listener/GroupUserRemovedListener.php
```diff
@@ -0,0 +1,34 @@
...
+
+	public function handle(Event $event): void {
+		if (!($event instanceof UserRemovedEvent)) {
+			return;
+		}
+        
+        // Get all shared albums for this group:
+        $albums_group = $this->albumMapper->getSharedAlbumsForCollaboratorWithFiles($event->getGroup()->getGID(), AlbumMapper::TYPE_GROUP);
+        // Get all albums shared with this specific user:
+        $albums_user = $this->albumMapper->getSharedAlbumsForCollaboratorWithFiles($event->getUser()->getUID(), AlbumMapper::TYPE_USER);
```
artonge, line 25: Same:

- You have the list of albums for a group
- For each album:
    - If the use is not in the collaborator list, remove its files.

lib/Listener/GroupDeletedListener.php
simonspa: > * If the use is not in the collaborator list, remove its files.

This is the culprit - in order to know if the user is collaborator of a given album, I need to fetch this - so one more DB query for every album. Whether I loop over users or albums doesn't matter.

Am I missing something obvious? :thinking:

lib/Listener/GroupDeletedListener.php
```diff
@@ -0,0 +1,48 @@
...
+	public function __construct(AlbumMapper $albumMapper) {
+		$this->albumMapper = $albumMapper;
+	}
+
+	public function handle(Event $event): void {
+		if (!($event instanceof GroupDeletedEvent)) {
+			return;
+		}
+        
+        // Get all shared albums for this group:
```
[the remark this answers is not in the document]
simonspa: Arg, my editor doesn't like me and switched from tab indentation to spaces... :/ Fixed now.

lib/Listener/GroupDeletedListener.php
```diff
@@ -0,0 +1,48 @@
...
+        $albums_group = $this->albumMapper->getSharedAlbumsForCollaboratorWithFiles($event->getGroup()->getGID(), AlbumMapper::TYPE_GROUP);
```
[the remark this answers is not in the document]
simonspa: Good idea, done that. :+1:

lib/Listener/GroupDeletedListener.php
artonge: Ah, I thought that the list of collaborators could be retrieved from `AlbumInfo`. But it is not. Sorry for the trouble !

artonge: In this case, adding a new method `getAlbumsForCollaboratorId` to `AlbumMapper` is necessary to at least reduce the DB load.

lib/Listener/GroupDeletedListener.php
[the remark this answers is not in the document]
artonge: Can't see it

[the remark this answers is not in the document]
simonspa: Odd, but must be a Github interface issue, it's here: https://github.com/nextcloud/photos/pull/1601/commits/c2ac203888f05082742ccfda2f8959b03cd3c1db

lib/Listener/GroupDeletedListener.php
simonspa: I agree, and that's done as suggested by you above. Not sure why it doesn't show here.

Issue #1591, opened 2023-01-16
AxiosError: Request failed with status code 404 when deleting photo


andyxheli:
**Describe the bug**
This only happens when deleting the photo while having it open. 

**To Reproduce**
1. Go to 'Photos'
2. Click on a Photo
3. Select 3 dot menu
4. Select Delete

NC 25.0.2
App V 2.0.1

![image](https://user-images.githubusercontent.com/59488153/212760527-f92547b1-dbdf-4c3f-9b9c-3cf036b77fe8.png)

![image](https://user-images.githubusercontent.com/59488153/212760715-7c01e2b0-d45a-4343-9582-85014e4703dc.png)

simonspa:
Somehow the path for downloading (menu item above) is correct (apart from a double slash, e.g. `http://localhost:8210/remote.php/dav//files/admin/Media/photo-1503991721143-75f95ebf1e55.jpeg`), but the endpoint for deletion is wrong:

```
DELETE http://localhost:8210/remote.php/dav/files/admin/files/admin/Media/photo-1503991721143-75f95ebf1e55.jpeg
```

containing the base path twice (`files/admin/files/admin`)

Is this an issue in photos or in https://github.com/nextcloud/viewer?

chw90:
I can confirm this issue on NC 25.0.2 when trying to delete an opened photo as described from the user's web interface.

nrosado:
I have the same problem. I'm on version 25.0.3

nrosado:
> 

I'm using an Oracle Free Pier to the main installation. Them my gold is to get a connection to my Synology NAS to do the backups. Don't know if the issues from the type of environment of Oracle or an issue on Nextcloud configuration. I'm not a pro to understand that.

Dbroek:
i have the same issue. I am using Nextcould behind a reverse proxy. (Nignx)
downloading files is no problem, jsut deleting gives this issue.

mitsol:
Same here, also behind a reverse Proxy

umgfoin:
[...]

PR #1602, opened 2023-01-22
Fix viewer in public albums

### LINKED ISSUE #1440: Photos that are shared by another user in a collaborative album display an error in the viewer
**Describe the bug**
Photos that are shared by another user in a collaborative album display an error in the viewer. (does not work both ways: User can only view their own respective photos in a collaborative album with the viewer)

**To Reproduce**
1. Create and run a nextcloud 25 docker image
2. Create a second user "Test" next to the user "admin"
3. Create a Album with the admin account, and add two images (nextcloud's samples images)
4. Share the Album with user Test
5. Login with user Test and add a photo to the collaborative album (e.g., one of nextcloud's sample images)
6. Login again with admin and view the added photo with the viewer

**Expected behavior**
The shared image should be viewable with nextcloud's viewer

**Screenshots**
Screenshot from 2022-10-29 12-15-09
Screenshot from 2022-10-29 12-16-39

**Browser log**
```
Error loading /photos/testuser/sharedalbums/Testalbum (admin)/11-Testalbum (admin) 
error { target: img.viewer__file--hidden.viewer__file, isTrusted: true, srcElement: img.viewer__file--hidden.viewer__file, eventPhase: 0, bubbles: false, cancelable: false, returnValue: true, defaultPrevented: false, composed: false, timeStamp: 718513, … }
Mime.js:111:10
```

### LINKED ISSUE #1452: [Bug] Viewer in public albums is broken
**Describe the bug**
When clicking on a thumbnail in public album view, the image is not loaded, but an error message "Error loading ..." is displayed.

The console reveals that the app is trying to load something like

https://nextcloud.example.com/apps/files_sharing/publicpreview/null?file=/photospublic/G3gmrpTgqfRLI1GA6RRG1yoyLoIhGyVM/275392-myimage.jpg&fileId=275392&x=2560&y=1440&a=true

which returns a 404 error.

**To Reproduce**
1. Create an album
2. Create a public link
3. Open the link (in a fresh browser / incognito mode)
4. Click on a thumbnail

**Expected behavior**
The image should be displayed

**Screenshots**

**Server:**
 - Nextcloud 25.0.1

starypatyk:
This is an attempt to fix #1452 (and related duplicates). It consists of two related changes.
Also fix #1440 

**Part 1** - commit https://github.com/nextcloud/photos/commit/1d1f0110de0d8ef53140699534291cd42ddc01d6

If I understand correctly, the WebDAV requests to `/remote.php/dav/photospublic/album-token/id-filename` are supposed to be available anonymously - for everyone who has a link to the album, without a need to authenticate first.

The `Sabre\DAV\Auth\Backend\AbstractBasic` class expects the `Authorization: Basic` header, and returns 401 error code if it is not present:
https://github.com/sabre-io/dav/blob/master/lib/DAV/Auth/Backend/AbstractBasic.php#L101

I think that in this case we should just allow anonymous access - by implementing the `Sabre\DAV\Auth\Backend\BackendInterface` interface directly. The `PublicAlbumAuthBackend` class is registered only for DAV URLs starting with `photospublic`, so it seems safe to always report success in the `check` method.

See: https://github.com/nextcloud/photos/blob/master/lib/Listener/SabrePluginAuthInitListener.php#L47

**Part 2** - commit https://github.com/nextcloud/photos/commit/166c63530fb18b4e58996a20c907bc2f955d08f2

I propose two changes in `PublicAlbumContent.vue`:

- Set `hasPreview` to `false` for all files in the album. This will force **current** implementation in the **Viewer** app to use the source URL, rather than the default preview URL that does not work in this case. These changes alone should suffice to fix the issue - at least temporarily.
- The second change - setting the `appPreviewPath` - is a proposal how the app can provide a custom preview URL to the **Viewer** app. This is obviously not yet handled - I plan to submit a complementary PR to the **Viewer** app.

Comments/suggestions?


simonspa:
@starypatyk thanks a lot, I'll test in a bit!

I noticed that also in "regularly" shared albums there is an issue whenever trying to open a photo from another user in the viewer, also seems related to a wrong endpoint.

...and one more: the deletion in Viewer has a wrong path: https://github.com/nextcloud/photos/issues/1591

Any chances these are also related to the same auth mixup?

starypatyk:
@simonspa 

> @starypatyk thanks a lot, I'll test in a bit!

Thanks! :+1: 

> I noticed that also in "regularly" shared albums there is an issue whenever trying to open a photo from another user in the viewer, also seems related to a wrong endpoint.
> 
> ...and one more: the deletion in Viewer has a wrong path: #1591
> 
> Any chances these are also related to the same auth mixup?

Most likely these are unrelated. My changes are relevant only for public albums.

simonspa:
Works like a charm for me, now clicking on thumbnails in a publicly shared album correctly opens the image in the viewer. Also downloading the original from there works.

When removing the public share and accessing the file DAV endpoint, i.e. `http://localhost:8210/remote.php/dav//photospublic/oMtN1RGlMrJvNdCheowrZ9XNuVqg4obD/413-IMG_8480.jpg` I get a full page returned with:

Screenshot from 2023-01-22 18-51-35

Shouldn't this just be a HTTP error code?

simonspa (commented):
Right now there is no bruteforce checking anymore, right? So even if an album link does not lead to a correct album returned this is not registered.

Could we add this back?

starypatyk:
@simonspa 

> When removing the public share and accessing the file DAV endpoint, i.e. `http://localhost:8210/remote.php/dav//photospublic/oMtN1RGlMrJvNdCheowrZ9XNuVqg4obD/413-IMG_8480.jpg` I get a full page returned with:
> 
> Screenshot from 2023-01-22 18-51-35
> 
> Shouldn't this just be a HTTP error code?

I think that this is the content that gets returned together with 404 error code. Apparently the output depends on the request - most likely based on the **Accept** header. When invoking an invalid URL with Curl I get (simplified for clarity):

```
HTTP/1.1 404 Not Found
Server: nginx
Date: Sun, 22 Jan 2023 18:07:07 GMT
Content-Type: application/xml; charset=utf-8
Content-Length: 219

<?xml version="1.0" encoding="utf-8"?>
<d:error xmlns:d="DAV:" xmlns:s="http://sabredav.org/ns">
  <s:exception>Sabre\DAV\Exception\NotFound</s:exception>
  <s:message>Unable to find public album</s:message>
</d:error>
```

starypatyk:
@simonspa 

> Right now there is no bruteforce checking anymore, right? So even if an album link does not lead to a correct album returned this is not registered.
> 
> Could we add this back?

Indeed - I have removed it. Some thoughts:

- Given that the token seems to be 192 bits, brute force attack is quite unlikely to be successful. :wink: 
- I think we should not include it in the `PublicAlbumAuthBackend` (to avoid wrong error codes like 401), maybe rather in `PublicRootCollection` - e.g. here: https://github.com/nextcloud/photos/blob/167b532734d2d1599aa73b4bd4a822cb408dbc88/lib/Sabre/PublicRootCollection.php#L82

simonspa:
> I think that this is the content that gets returned together with 404 error code.

Alright then. :) I now realize I get the same return also for regularly shared files. :+1: 

> brute force attack is quite unlikely to be successful

That's right, just not sure if it might reduce resource usage if we throttle early instead of querying the DB for every supposed album. But I really have no clue :smile:

starypatyk:
> > brute force attack is quite unlikely to be successful
> 
> That's right, just not sure if it might reduce resource usage if we throttle early instead of querying the DB for every supposed album. But I really have no clue smile

Me neither... :wink: I have not yet figured out how the throttling works. Can try to add it back if suggested.

marcelklehr:
I think having brute force protection on share endpoints would be good. here's the docs on that https://docs.nextcloud.com/server/latest/developer_manual/basics/controllers.html#brute-force-protection

starypatyk:
I have aligned the changes in `PublicAlbumContent.vue` to the changes done in the **Viewer** app discussed in the related PR: https://github.com/nextcloud/viewer/pull/1525

I had to revert setting the `hasPreview` property to `false` as this would break the proposed solution in the Viewer app.

This would need to be discussed further, if we would like to have this workaround available for currently released and older versions of the Viewer app.

starypatyk:
> I had to revert setting the `hasPreview` property to `false` as this would break the proposed solution in the Viewer app.
> 
> This would need to be discussed further, if we would like to have this workaround available for currently released and older versions of the Viewer app.

After https://github.com/nextcloud/viewer/pull/1525/commits/4a080e4257a49079e23091ac86640b7d49326844 the `previewUrl` has precedence over `hasPreview`. So it has been possible to once again disable default preview for public albums: https://github.com/nextcloud/photos/pull/1602/commits/d72624d33188efeef70a514a3c7f411fc218e4aa

szaimen:
fix in viewer was merged and backported

starypatyk:
@marcelklehr 

> I think having brute force protection on share endpoints would be good. here's the docs on that https://docs.nextcloud.com/server/latest/developer_manual/basics/controllers.html#brute-force-protection

In https://github.com/nextcloud/photos/pull/1602/commits/8d5572bbd47e1c6b10d0d126aed1daf6e43c58ef I have put brute force throttling back. As this needs to protect WebDAV requests, this seemed to me like a reasonable place.

This is very similar to the implementation here:
https://github.com/nextcloud/server/blob/41148acf833d401aa6c8bd23617ae8639b6aaae6/apps/dav/lib/Connector/PublicAuth.php#L79

But I am not satisfied with the result. :disappointed: 

I have tried to open an invalid URL a few times and the throttling turned on very quickly - I started to get significant delays (10 seconds or more). Unfortunately, this continued even when I finally used a correct URL.

I imagine this might easily happen for regular users - in situations like missing last character when copying a link, someone enabling/disabling sharing of the same album, etc.

Might be reasonable to change the throttling parameters to make it less intrusive. Or maybe resetting throttling when the user provides a correct link.

Thoughts/suggestions?

artonge:
@starypatyk can you check whether you can implement it as suggested by the doc ? https://docs.nextcloud.com/server/latest/developer_manual/basics/controllers.html#rate-limiting
Then you can configure it to something like `(limit=10, period=60)` which would probably be enough.

starypatyk:
@artonge @marcelklehr

I have looked at the code in `OC\Security\Bruteforce\Throttler` and `OC\Security\RateLimiting\Limiter` that implement the underlying brute force throttling and rate limiting functions.

**TL;DR** - I would suggest **not to implement** either of these mechanisms for public albums.

Here is my reasoning:

1. First of all - these links are intended to be public, so general throttling when the user provides a **correct** address (i.e. token) should not be applied.
2. Since the tokens are long (192 bits apparently) brute-forcing is highly unlikely to succeed.
3. Current implementation of brute force throttling does not allow to configure throttling parameters. With current parameters the mechanism might be quite annoying when a user makes some simple mistake.
4. The rate limiting mechanism is completely separate from the brute force throttling and does not provide any delay. When the rate is exceeded it just raises an exception.
5. The rate limiting mechanism logs additional information in the database to determine the rate of incoming requests.
6. Due to item #1 above - we **must** verify the token before even trying to involve the rate limiting mechanism.
7. So - if we add the rate limiting mechanism after verifying the token, we will add additional database load, and the outcome will just be a different kind of exception if the rate is exceeded. We already raise an exception and terminate processing of the request when the token is invalid.
8. **Conclusion** Leaving brute force throttling might be annoying. Adding rate limiting would increase load on the server (processing and database) without giving any benefits.

If you agree with this, I will revert the last commit: https://github.com/nextcloud/photos/pull/1602/commits/8d5572bbd47e1c6b10d0d126aed1daf6e43c58ef

It is possible to implement a more advanced rate limiter that would also introduce a delay and hopefully reduce the server load. But this seems to be a topic for another PR - likely in the core server code. :wink:

lib/Sabre/PublicAlbumAuthBackend.php
```diff
@@ -21,46 +21,58 @@
...
-		$albums = $this->albumMapper->getSharedAlbumsForCollaboratorWithFiles($username, AlbumMapper::TYPE_LINK);
-
-		if (count($albums) !== 1) {
-			$this->throttler->registerAttempt(self::BRUTEFORCE_ACTION, $this->request->getRemoteAddress());
-			return false;
-		}
-
-		\OC_User::setIncognitoMode(true);
+	public function check(RequestInterface $request, ResponseInterface $response) {
+		\OC_User::setIncognitoMode(true); // ???
```
artonge, line 59: From the declaration: `set incognito mode, e.g. if a user wants to open a public link`

```suggestion
		\OC_User::setIncognitoMode(true);
```

lib/Sabre/PublicAlbumAuthBackend.php
```diff
@@ -21,46 +21,58 @@
...
+		return [true, "principals/token"];
```
artonge, line 60: I understand that we need to replace 'token' with the actual token in the returned value.

```suggestion
	public function check(RequestInterface $request, ResponseInterface $response) {
		\OC_User::setIncognitoMode(true);

		$auth = new Sabre\HTTP\Auth\Basic(
			$this->realm,
			$request,
			$response
		);

		$userpass = $auth->getCredentials();
		if (!$userpass) {
			return [false, "No 'Authorization: Basic' header found. Either the client didn't send one, or the server is misconfigured"];
		}

		return [true, "principals/".$userpass[0]];
	}
```

artonge (changes requested):

lib/Sabre/PublicAlbumAuthBackend.php
starypatyk: I am still not 100% sure what the `setIncognitoMode` call actually does. I suspect that it prevents using information about the logged-in user in further processing. However, I do not know how this applies in this case.

lib/Sabre/PublicAlbumAuthBackend.php
starypatyk: I am not sure what is the meaning of the return value in this case. I have copied the comment above the `check` method from the original `Sabre\DAV\Auth\Backend\BackendInterface` file: https://github.com/sabre-io/dav/blob/master/lib/DAV/Auth/Backend/BackendInterface.php#L19
I suspect that the returned value is not used at all. Only when the `check` method returns `[false, "error msg"]` the request is rejected.

Actual token verification is performed in `PublicRootCollection`:
https://github.com/nextcloud/photos/blob/4c87d2566bdf405bb5f430cc8fd839be64587064/lib/Sabre/PublicRootCollection.php#L92

Regarding bringing back the `Sabre\HTTP\Auth\Basic` class and the `getCredentials` call - this is exactly what I wanted to avoid when suggesting changes to the `PublicAlbumAuthBackend`. We **want the request to succeed** for anonymous users - i.e. when there is no `Authorization: Basic` header. This was the reason why the original code failed when an anonymous user tried to download pictures from a shared album.

artonge: You're right, let's keep your version :)

artonge (approved):

szaimen:
@artonge is cypress failure related?

artonge:
I don't think so. Let's rerun, we'll see

szaimen:
Thanks a lot @starypatyk for fixing this and @simonspa for the help and @artonge for reviews! :)

starypatyk:
@szaimen @artonge @marcelklehr 

Just one thing - I have **not** reverted the brute force throttling introduced in commit https://github.com/nextcloud/photos/pull/1602/commits/7922c6977f5b4100f19a8ea9d1be959043a8ee07 as discussed in the comment above: https://github.com/nextcloud/photos/pull/1602#issuecomment-1414386749, as I was not sure if it is OK to do so.

Should I submit a separate PR to discuss this?

artonge:
> Should I submit a separate PR to discuss this?

I think it is better to keep it for now.

PR #1611, opened 2023-01-25
fix(tags performance): Cache tag counts on server and load them using initial state

marcelklehr:
... and only fetch file lists when the corresponding tag is visible

This should greatly improves UX and performance of the tags view. Before, the popular tags would continuously jump around for dozens of seconds while the tag counts were being loaded from the server.


lib/Controller/PageController.php
```diff
@@ -125,14 +137,28 @@ public function index(): TemplateResponse {
...
+		$key = $user->getUID() . ':' . $userFolder->getEtag();
+		$tagCounts = $this->tagCountsCache->get($key);
+		if ($tagCounts === null) {
+			$tags = $this->tagManager->getAllTags(true);
+			$tagCounts = [];
+			foreach ($tags as $tag) {
+				$search = $userFolder->search(new SearchQuery(new SearchComparison(ISearchComparison::COMPARE_EQUAL, 'systemtag', $tag->getName()), 0, 0, [], $user));
+				$tagCounts[$tag->getName()] = count($search);
+			}
+			$this->tagCountsCache->set($key, $tagCounts, 60 * 60 * 24 * 7); // 7 days
```
artonge, line 158: Shouldn't this be invalidated when the number of item changes ?
I am not familiar enough with `ICache`.

src/views/Tags.vue
```diff
@@ -26,7 +26,7 @@
 		<!-- Errors handlers-->
 		<NcEmptyContent v-if="error" :title="t('photos', 'An error occurred')" />
 
-		<NcEmptyContent v-if="!loading && !hasTagsWithFiles" :title="t('photos', 'No tags yet')" :description="t('photos', 'Photos with tags will show up here')" />
+		<NcEmptyContent v-if="!loading && !tagsList.length" :title="t('photos', 'No tags yet')" :description="t('photos', 'Photos with tags will show up here')" />
```
artonge, line 29:
```suggestion
		<NcEmptyContent v-if="!loading && tagsList.length === 0" :title="t('photos', 'No tags yet')" :description="t('photos', 'Photos with tags will show up here')" />
```

lib/Controller/PageController.php
marcelklehr: Yeah, I'm not super confident about this bit. I found it important to add caching. Currently it will invalidate based on the user folder etag, but I'm not sure if that is affected by tags. Improvements welcome!

artonge: How do you retrieve the list of files for a tag ? maybe we could invalidate there too?

marcelklehr: The client fetches files [here](https://github.com/nextcloud/photos/blob/master/src/services/TaggedImages.js). After the file list has been fetched for a tag on the client (which only happens if the tag is visible), the cached tagCount for that tag is overridden with the actual length of the file list. The point of this initial state (and cache) is to avoid having to wait for all tags' file lists to have been fetched until the top 10 tags can be displayed. Thus, if this cache entry is slightly outdated, it shouldn't matter too much, I think.

cc @skjnldsv What do you think?

skjnldsv: Sorry, I do not have the space for such a deep context switch :/

artonge: Ah yes, it is a DAV request on an endpoint not controlled by photos.
Either listen to https://github.com/nextcloud/server/blob/ecad09b38a152957c771139562e61f7ecb1a6879/lib/public/SystemTag/MapperEvent.php#L37

Or simply recreate the cache every 30 min or so ? Is the search taking a long time ?

artonge: If the search is fast, maybe we do not need to cache it.

marcelklehr: Listening to a tag event is probably the best approach.

artonge (approved):

marcelklehr:
/backport 8ba4c1e728d1ed80777febc3773841d263160747,ba7998f05543afa78f87600b8b781bf08b5a8389,dee1b276b29a64a4c2dfbe07ef4467a809203a7f to stable25

PR #1621, opened 2023-02-01
Consolidate fontend logic

artonge:
This consolidates the frontend store and fetcher logic for album, shared albums and public albums.
This uses the work done on the locations views https://github.com/nextcloud/photos/pull/1334

- Unify how the store is handled
- Unify how the collections are fetched

Hopefully, this makes the code easier to maintain and contribute.

This could be extended to tag and faces, but I know Marcel did some other improvements/tweaks there, so keeping it to album for now.


Fenn-CS (approved):
Looks okay to me.

_The earlier, we remove this compiled files from tracking I think the better!_

PR #1634, opened 2023-02-10
Create headings for photos pages

JuliaKirschenheuter:
- [x] Blocked by https://github.com/nextcloud/server/pull/36640

Relates to https://github.com/nextcloud/server/issues/35626, fixes "Activity" part

After:

Screenshot from 2023-02-13 15-34-00


ChristophWurst (approved):

lib/Controller/PageController.php
```diff
@@ -166,7 +166,9 @@ public function index(): TemplateResponse {
 			Util::addTranslations('recognize');
 		}
 
-		$response = new TemplateResponse(Application::APP_ID, 'main');
+		$response = new TemplateResponse(Application::APP_ID, 'main', [
+			'pageTitle' => 'Photos'
```
Pytal, line 170: Would suggest that we do not add variables that aren't used, thoughts @artonge?

Pytal: Actually it's used in the server templates https://github.com/nextcloud/server/pull/36640 so please ignore :)

Pytal (approved):

lib/Controller/PageController.php
artonge, line 170: No translation ?

artonge (approved):

Issue #1394, opened 2022-10-19
"Download all files" produces an empty TAR file


nicokaiser:
**Describe the bug**
The "Download all files" feature does not work. An empty (1024 bytes, no contents) "download.tar" file is downloaded.

**To Reproduce**
1. Open Photos app
2. Navigate to any Album
3. Open the "..." menu
4. Click on "Download all files in album"

**Expected behavior**
A ZIP (not TAR) file should be downloaded with all files in the album.

**Desktop (please complete the following information):**
 - OS: macOS
 - Browser: Chrome

**Server:**
Nextcloud 25.0.0

I am using the Dockerfile from the nextcloud/docker master branch. Maybe this is fixed with the official Docker image, I will check this once it is uploaded to Docker Hub.

skjnldsv:
@artonge :thinking: 

Btw @nicokaiser , thank yuo so much for your tickets!! :pray:

nicokaiser:
> Btw @nicokaiser , thank you so much for your numerous tickets!! It truly helps 🙏

I am eagerly waiting for the Photos 2.0 app (not updated my production server to NC25), so I'm more than happy to help making it even more awesome! Especially the possibility to share albums (and make public links from albums) will be a very useful feature...

Raemsna:
I confirm and 2nd this issues. 
See link above with additional information

szaimen:
Can yoi try again with 25.0.1?

nicokaiser:
Still the same with 25.0.1.

BkSouX:
same issue here with Nextcloud Hub 3 (25.0.2) and Photos 2.0.1

oswaldo-be:
and unfortunately really annoying when you think for hours it is a problem of the nextcloud itself or another component

artonge:
The button was removed by https://github.com/nextcloud/photos/pull/1641 until a proper way to download albums' content comes

jonesints:
I think the original bug report ""Download all files" produces an empty TAR" is related to library TarStreamer:

https://github.com/owncloud/TarStreamer

php8 isn't supported and I believe owncloud had a different roadmap regarding php8 support which might be causing this issue:

[https://github.com/owncloud/TarStreamer/issues/23#issue-1155236540](url)

[https://central.owncloud.org/t/owncloud-development-roadmap-future-of-owncloud-classic-versus-ocis/37633/2](url)

[...]

PR #1641, opened 2023-02-21
Disable download from albums

artonge:
Temporary fix for https://github.com/nextcloud/photos/issues/1394


szaimen:
@artonge I suppose the correct fix would be much work?

artonge:
> @artonge I suppose the correct fix would be much work?

Yes, sadly, it would involve a plugin system to download files, as the current download API is not able to check whether a file is accessible from an album.

PR #1652, opened 2023-02-24
fix(psalm): psalm fix

skjnldsv:
Signed-off-by: John Molakvoæ <skjnldsv@protonmail.com>

### DIFF
--- a/lib/Controller/PreviewController.php
+++ b/lib/Controller/PreviewController.php
@@ -129,7 +129,12 @@ class PreviewController extends Controller {
 	}
 
 
-	protected function getFileIdForAlbums($fileId, $albums) {
+	/**
+	 * @param (\OCA\Photos\Album\AlbumInfo|mixed)[] $albums
+	 *
+	 * @psalm-param array<\OCA\Photos\Album\AlbumInfo|mixed> $albums
+	 */
+	protected function getFileIdForAlbums(int $fileId, array $albums) {
 		foreach ($albums as $album) {
 			$albumFile = $this->albumMapper->getForAlbumIdAndFileId($album->getId(), $fileId);
 			$nodes = $this->rootFolder
--- a/lib/Sabre/PhotosHome.php
+++ b/lib/Sabre/PhotosHome.php
@@ -95,7 +95,9 @@ class PhotosHome implements ICollection {
 	}
 
 	/**
-	 * @return (AlbumsHome)[]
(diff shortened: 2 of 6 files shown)


lib/Controller/PreviewController.php
```diff
@@ -129,7 +129,12 @@ public function index(
 	}
 
 
-	protected function getFileIdForAlbums($fileId, $albums) {
+	/**
+	 * @param (\OCA\Photos\Album\AlbumInfo|mixed)[] $albums
+	 *
+	 * @psalm-param array<\OCA\Photos\Album\AlbumInfo|mixed> $albums
```
artonge, line 135: What is the difference ?

lib/Sabre/PhotosHome.php
```diff
@@ -95,7 +95,9 @@ public function getChild($name) {
 	}
 
 	/**
-	 * @return (AlbumsHome)[]
+	 * @return (AlbumsHome|PlacesHome|SharedAlbumsHome)[]
+	 *
+	 * @psalm-return array{0: AlbumsHome, 1: SharedAlbumsHome, 2: PlacesHome}
```
artonge, line 100: Same, what is the difference ? Is `@psalm-return` allowing for a more fine grain type description ? If so, is `@return` still needed ?

lib/Sabre/Place/PlaceRoot.php
```diff
@@ -83,7 +83,9 @@ public function createDirectory($name) {
 	}
 
 	/**
-	 * @return PlacePhoto[]
+	 * @return (PlaceFile|PlacePhoto)[]
```
artonge, line 86: I think it only returns `PlacePhoto`.
Those names are really confusing :see_no_evil:

lib/Controller/PreviewController.php
skjnldsv: This is an automated psalm fix, it knows better than me :see_no_evil:

lib/Sabre/Place/PlaceRoot.php
skjnldsv: If psalm found the two, then something is wrong :wink:

artonge: Ah, it is because of the declaration of `$children` !
Fixed :)

artonge (approved):

PR #1664, opened 2023-02-28
Further optimize preview rendering

artonge:
- Delegate loading priority to the browser
- Check cache if previews are already loaded

Tip: easier to review per commit.


src/components/File.vue
```diff
@@ -33,27 +33,41 @@
...
+						ref="imgLarge"
+						:key="`${file.basename}-large`"
+						:src="srcLarge"
+						:alt="file.basename"
+						:decoding="loadedLarge || isVisible ? 'sync' : 'async'"
+						:fetchpriority="loadedLarge || isVisible ? 'high' : 'low'"
+						:loading="loadedLarge || isVisible ? 'auto' : 'lazy'"
+						:aria-describedby="ariaDescription"
+						@load="onLoadLarge"
+						@error="onErrorLarge">
```
skjnldsv, line 65: Looks like there might be room for a dedicated component or unifying those lines? :)

artonge: I which, but they are slightly different and also dependent on each other, so an abstraction layer would bring unnecessary complexity. Also, in the current state, it is easily tweakable, and easier to read.

skjnldsv (approved):

skjnldsv:
If you have any good tip for Files after your expertise with this PR, I'll take them :pray: :bow:

AndyScherzinger:
@artonge has this also been backported to 25?

artonge:
No, and looks like I also did not backport to 26. I probably considered this too big of a change to be backported, and it is also not a bug fix.
If really needed, I can backport.

Issue #1540, opened 2022-12-15
Not all subfolders are displayed depending on the scale of the page


skl256:
**Describe the bug**
Not all subfolders are displayed depending on the scale of the page.
For example, in the folder "2021" we have 143 subfolders.
The Files app shows all 143 subfolders without issue.
In the Photos app, when you open the 2021 folder, only 77 subfolders are displayed (1920*1080 screen, 100% page zoom).
If you start zooming out of the page in the browser, then the missing folders will appear below.

**To Reproduce**
1. Create a folder.
2. Put there, for example, 150 folders, each of them should contain photos.
3. Go to NEXTCLOUD_DOMAIN/apps/photos/ in the left menu, select "Folders", open a folder that contains 150 photo folders inside.
4. Scroll down, you will see that not all folders are displayed.
5. "Play" with the scale of the page, you will see that the number of visible folders changes.

**Screenshots**
![image](https://user-images.githubusercontent.com/111241000/207895015-1337b9c7-8eeb-4215-8144-645d654204a9.png)

**Desktop (please complete the following information):**
 - OS: Windows 11
 - Browser: Chrome
 - Version  108.0.5359.100

**Additional context**
The problem also occurs on other browsers, including mobile devices.

NC: 25.0.2-fpm

medarion:
I can confirm this issue. It always happens for long lists of files/folders after scrolling a few pages. The scrollbar has the right length, so enough space is reserved, but the thumbnails are not displayed. Any action which changes the viewport size like zooming, resizing the browser or opening the developer sidebar causes the thumbnails to be loaded.
Im running Ubuntu and this happens both in Firefox and Chromium (recent versions).

JoshuaPettus:
Confirmed as well on firefox. resizing does effect which photos or folders are loaded, but it still wont display all of them at once for me.  I just get a bunch of blank space before and after that never loads.    I hope this one can get fixed soon

temotodochi:
Confirmed on firefox.

htttito:
Confirmed on firefox

joshtrichards:
A fix for this was made in #1695 and backported as far back as v25.0.6. Can you confirm it is fixed in your environment? If so, please close this Issue!

medarion:
I can confirm the issue is fixed in 27.0.0.

skl256:
Thanks @joshtrichards ! 
I check in my environment in 25.0.8 - fixed.
[...]

Issue #1490, opened 2022-11-21
Folder view only shows first 126 images and "Failed - No file" error


XXDoMi77:
**Describe the bug**
Folder view only shows first 126 images, once an image is opened howewer, it does allow me to go past this limit, even though the images aren't visible. Also trying to download any image in folder view, results in a "Failed - No file" error in chrome.

**To Reproduce**
1. Go to folder view
2. Open a folder with more than 126 images
3. See error

**Desktop:**
 - OS: windows, android
 - Browser: chrome, edge
 - Photos version: 2.0.0

![Animation6](https://user-images.githubusercontent.com/27887974/203174580-14bb3c00-e0da-4dbd-afcf-51928f17ff11.gif)

skl256:
https://github.com/nextcloud/photos/issues/1540
With folders some problem too in Photos app ;(

joshtrichards:
A fix for this was made in #1695 and backported as far back as v25.0.6. Can you confirm it is fixed in your environment? If so, please close this Issue!

PR #1695, opened 2023-03-09
Fix scrolling in large folders

starypatyk:
When a folder contains over ~100 files or sub-folders (exact number seems to depend on the browser windows size), only a first batch of files/sub-folders is displayed. Empty space is shown when scrolling further down. This has been described in #1540 and likely also in #1490.

This seems to be the same issue as the one described in the **vue-virtual-grid** repository: https://github.com/Mikescops/vue-virtual-grid/issues/25

The fix provided there (https://github.com/Mikescops/vue-virtual-grid/pull/26) introduces a new property: `scrollElement` that should be set to an appropriate element on the page - i.e. the scrolling container.

In this PR I propose a fix implemented similarly to what is used e.g. in `FaceContent.vue`: https://github.com/nextcloud/photos/blob/65d222db10ab5f005125d56576b8bd223b4dd054/src/views/FaceContent.vue#L127 https://github.com/nextcloud/photos/blob/65d222db10ab5f005125d56576b8bd223b4dd054/src/views/FaceContent.vue#L251


artonge (approved):
If this fix the issue in your testing, then it looks fine :)

starypatyk:
Hi @artonge,

Great to see the fix merged into **master**. :+1:

Would you consider backporting it to **stable26** and **stable25**?

The change applies cleanly to both these branches (except compiled JS of course). Pre-requisites are also there - i.e. the required element on the page and appropriate version of the vue-virtual-grid library.

_I suspect I could invoke the nextcloud-bot myself, but I am not sure if this is an accepted practice for external contributors._

artonge:
> _I suspect I could invoke the nextcloud-bot myself, but I am not sure if this is an accepted practice for external contributors._

Stumbling on that, and it looks like we did not answer you. Feel free to use the backport bot next time if you feel like it is needed. In any case, the generated PRs will need to go through another round of approval ;).

PR #1699, opened 2023-03-12
[stable26] Update nextcloud/ocp dependency

nextcloud-command:
Auto-generated update of [nextcloud/ocp](https://github.com/nextcloud-deps/ocp/) dependency


artonge:
Done in https://github.com/nextcloud/photos/pull/1736

PR #1718, opened 2023-03-30
albums: check copy source owner is the current user and throw

pulsejet:
The current behavior when trying to copy a shared file to an album is meaningless. Since the `copyInto` call returns `false`, DAV falls back to trying to create the file again with the same content (see [this](https://github.com/sabre-io/dav/blob/8e09bff0d50ecdc3c68b3dffa59302c1d57d5f9d/lib/DAV/Tree.php#L132-L134) fallback). As a result,

1. The shared file gets copied over to the current user.
2. It gets added to the album.
3. The call fails anyway because the name of the destination node (that is copied inside the album) has changed, so the subsequent call to get the node from the album fails.

If we throw while copying, the user can't add shared files to albums anymore with a proper error message, and files don't get copied unnecessarily.

One thing to note here, I do think ideally there should be some way to add shared files to albums, as long as the file as resharing permissions. Of course then we might need to check later if the file is unshared and remove it from the album. Thoughts?


lib/Sabre/Album/AlbumRoot.php
```diff
@@ -156,13 +156,17 @@ public function getLastModified(): int {
...
-			return $this->addFile($sourceId, $ownerUID);
+		if (!$sourceNode instanceof File) {
+			throw new Forbidden("The source is not a file");
 		}
-		$uid = $this->userId;
-		throw new \Exception("Can't add file to album, only files from $uid can be added");
+
+		$sourceId = $sourceNode->getId();
+		$ownerUID = $sourceNode->getFileInfo()->getOwner()->getUID();
+		if ($ownerUID !== ($uid = $this->userId)) {
```
artonge, line 165: Could you do this assignment outside the condition ?

artonge (changes requested):
Indeed, much cleaner.
One small nitpick though

artonge:
> One thing to note here, I do think ideally there should be some way to add shared files to albums, as long as the file as resharing permissions. Of course then we might need to check later if the file is unshared and remove it from the album. Thoughts?

I don't remember why the limitation exists, but it was probably to reduce the complexity of the first version.
Could be a nice addition to support shared files :+1:

lib/Sabre/Album/AlbumRoot.php
pulsejet: Done

PR #1749, opened 2023-04-17
Do not allow deletion from public albums

artonge:
Time didn't allow for a fully editable public album, so let's remove this half-baked feature for now.


skjnldsv (approved):

iNETTechnik:
but still there...

Issue #1665, opened 2023-02-28
[BITV] 9.1.3.1b/4.2 - The "media" list was implemented semantically incorrectly - the "li" elements are not implemented as children of the "ul" element, but are inside "div" elements. (1)


AndyScherzinger:
<img width="657" alt="8dbb463b3c7828316ae50cb029cebd52" src="https://user-images.githubusercontent.com/1315170/221933486-f7894d46-03f2-4dcc-886c-dfd307d0768e.png">

![0e4660212b80e42b753b1adcc08a8370](https://user-images.githubusercontent.com/1315170/221933512-8958f61b-8975-440f-8209-018edc46cd7e.png)

<details>
<summary>Details</summary>

https://report.bitvtest.de/default-en/d63601ac-cb34-4645-8256-66bec78964a0.html#checkpoint-1d5eadb58d-v4-n2

</details>

JuliaKirschenheuter:
Dear @michaelnissenbaum,

I've looked into current implementation and changed the structure to the semantically right list but that solution breaks our layout ;(
I would like to ask you if it would be ok to create only container elements with `<div>` for photos and videos? Amazon photos and Google photos have implemented a media list this way:

<img width="1921" alt="3842-1882-max" src="https://user-images.githubusercontent.com/6078378/235942515-060990c7-79a8-4777-94bc-c035f7bd5288.png">

![Screenshot from 2023-05-03 16-15-05](https://user-images.githubusercontent.com/6078378/235942480-24e6217a-793a-4e29-8cfa-a662ca58dcb8.png)

Is there something more to pay attention on (probably some `aria-label`)?

michaelnissenbaum:
Hi @JuliaKirschenheuter. The two examples you sent me, from my perspective, are not truly accessible. The use of lists could theoretically be avoided, but in order to do so, the headings for each photo would need to be defined. As I see it, example number 2 demonstrates this possibility. However, we should only use headings when content is being overridden (i.e. further content follows). In our case, this is not the situation, so the only option remaining is to use lists.

AndyScherzinger:
@skjnldsv would you have an idea on how we could implement a mosaic view of photos (like we have) using a (ul-) list.

@michaelnissenbaum do you have any examples of a accessible implementation of such a gallery of images that follows the visual style of a mosaic-style?

JuliaKirschenheuter:
> @skjnldsv would you have an idea on how we could implement a mosaic view of photos (like we have) using a (ul-) list.

[...]

PR #1764, opened 2023-04-25
Create semantic correct list for the media list

JuliaKirschenheuter:
Resolves https://github.com/nextcloud/photos/issues/1665

Create a list for media list

|  Before   | After                                                        |
|-------------|---------------------------------------------------------------|
|Screenshot from 2023-04-25 16-18-21 | Screenshot from 2023-04-25 16-00-47


artonge (changes requested):
I recall that it was breaking the layout to not have a wrapper element per row.

Maybe it makes more sense to use a table instead of ul/li

JuliaKirschenheuter:
@AndyScherzinger @Pytal @artonge first implementation idea was regarding creating a list inside of a list for each title like: 
Screenshot from 2023-04-25 16-26-23

But I've tried several times and could not find a solution for it. For now i think it is a working solution and i will ask Michael if he could approve it.

artonge:
> Thanks Louis for your comment! I would like to know where layout could be broken, could you tell me some examples regarding it? Thanks!

It is a one in a hundred thing, but sometime, the rounding approximation makes the row too big and the last photo end up in the next row, creating cascading issues.

See the empty space in the second line of the screenshot:

Screenshot from 2023-04-25 16-43-20

JuliaKirschenheuter:
Thanks @artonge,
As i've implemented this task i've tried different approaches to reach the right list structure. And only this last solution have worked for current components implementation. 
@artonge and @Pytal do you have some other ideas how to reach the right list for current implementation environment?

jancborchardt:
> It is a one in a hundred thing, but sometime, the rounding approximation makes the row too big and the last photo end up in the next row, creating cascading issues.

@artonge If it only happens very very rarely, we could say it’s fine – or could we have an extra "buffer pixel" of sorts which prevents this rounding issue? That is, rows which would be too wide appear as 1px wider, but it’s totally fine and not really visible. Better than a whole image breaking.

To judge if it’s fine, @AndyScherzinger and I were wondering: Does the issue propagate into following lines, or will only the line with the issue look off? If only that line, it’s fine.

artonge:
> To judge if it’s fine, @AndyScherzinger and I were wondering: Does the issue propagate into following lines, or will only the line with the issue look off? If only that line, it’s fine.

It is cascading and happens too frequently to be acceptable. My memory was too generous, it's more once or twice out of every ten lines.

There is probably some way to fix that, but didn't find any yet.

artonge:
Superseeded by https://github.com/nextcloud/photos/pull/1764

PR #1766, opened 2023-04-27
Fixes #1694: Public album sharing broken

### LINKED ISSUE #1694: Album sharing broken
**Describe the bug**
Can't access shared album. Album shows in the list "shared albums" but if i try to access it i get "an error occured" on an empty Page.
error
error1
 On The second Album shared by another person i can infact open the album and see the previews, but enlarging the image brings up the following error:
error3

**To Reproduce**

see above - tried to reproduce instead got a different error...

**Expected behavior**
Access the  albums another person shared with me.

tested on Firefox110 and Chromium110 on Ubuntu 22.04 and on OSX
-> the second error seems to appear only with firefox.

umgfoin:
class OCA\Photos\Sabre\Album\PublicAlbumPhoto is not available.
Fixes #1694: Public album sharing aborts with error.

/compile amend /

### DIFF
--- a/lib/Sabre/Album/PublicAlbumRoot.php
+++ b/lib/Sabre/Album/PublicAlbumRoot.php
@@ -23,6 +23,7 @@ declare(strict_types=1);
 
 namespace OCA\Photos\Sabre\Album;
 
+include_once("PublicAlbumPhoto.php");
 use OCA\Photos\Album\AlbumFile;
 use Sabre\DAV\Exception\Forbidden;
 use Sabre\DAV\Exception\NotFound;


umgfoin:
For unknown reason,
autoload/ composer seems to ommit `lib/Sabre/Album/PublicAlbumPhoto.php`.
This leads to:
`Could not check compatibility between OCA\Photos\Sabre\Album\PublicAlbumRoot::getChild($name): OCA\Photos\Sabre\Album\PublicAlbumPhoto and OCA\Photos\Sabre\Album\AlbumRoot::getChild($name): OCA\Photos\Sabre\Album\AlbumPhoto, because class OCA\Photos\Sabre\Album\PublicAlbumPhoto is not available at [NC_ROOT]/apps/photos/lib/Sabre/Album/PublicAlbumRoot.php#82`
The only solution I found, is to explicitely pull in `PublicAlbumPhoto.php` from `PublicAlbumRoot.php`.

artonge:
> For unknown reason,

I think we should understand the reason all other files are properly loaded and not `PublicAlbumPhoto.php` as this should not be needed.

umgfoin:
Yes, preferrable.
I initially saw a problem in covariance of `PublicAlbumRoot::getChild($name) `and base class `AlbumPhoto::getChild($name) `
in conjunction with `declare(strict_types=1)` (Same for `::getChildren) `.
It didn't prove to be right - but I'm not really good in weakly typed languages.

artonge:
FYI, I highly suspect an issue on your side, as this is working fine in other places.

umgfoin:
> I highly suspect an issue on your side,

You're correct - problem found and resolved:
I had a git issue within the vendor-folder.

Issue #1714, opened 2023-03-27
UUID displayed when searching a user while sharing an album


marcelklehr:
**Describe the bug**
- [ ] UUID displayed when searching a user while sharing an album
- [ ] UUID also displayed in received shared albums

**To Reproduce**
1. Go to web app
2. Go to photo app
3. Go to "Album" menu
4. Share an album
5. Search either by firstname+lastname or mail address

**Expected behavior**
Should display the firstname and lastname

**Screenshots**
![ezgif-4-2eaa97a371](https://user-images.githubusercontent.com/986878/229489181-602461e4-1a58-470e-9da5-8ae3f788dd1c.gif)

**Desktop (please complete the following information):**
 - Browser: Chrome/Firefox/Opera/Safari

DanScharon:
can confirm with Nextcloud 25.0.5 / Photos 2.0.1

marcelklehr:
:tada: Thanks @artonge

PR #1781, opened 2023-05-10
Use display name instead of id for shared albums

artonge:
Fix https://github.com/nextcloud/photos/issues/1714

Before | After
-- | --
Screenshot from 2023-05-10 11-10-14 | Screenshot from 2023-05-10 11-10-40
Screenshot from 2023-05-10 11-10-22 | Screenshot from 2023-05-10 11-10-45
Screenshot from 2023-05-10 11-09-55 | Screenshot from 2023-05-10 11-09-26


skjnldsv (approved):

pulsejet:
I just want to point out that this potentially makes the name of the album ambiguous. For example, if two users with the same display name share an album with the same name to a user, then there's a conflict, and moving files to the album is ambiguous.

Notably, this could be an attack vector, since both the album name and display name can be changed by the user sharing the album. Of course there are non-malicious cases too, for example, two users named "Alice" sharing an album named "Christmas 2022" with Bob. The earlier behavior of using the username, on the other hand, was completely unambiguous.

@artonge @skjnldsv

artonge:
Ah yes indeed, that might have been why I used user ID back then. We could revert, but as userid is not really made for display, then we end up with potentially weird titles.

Other options could be to add both, display name then userid. Userid would be less visible. Or to add a title saying something like "Shared by userid".

We could also rethink how we list shared albums. Maybe having a section per users:

Display name 1 (userid1):
Album name 1      Album name 2     ....

Display name 2 (userid2):
....

nimishavijay:
Interesting problem! We generally don't use user IDs for display anywhere. There are cases even in other apps where this could happen, for example an event in a calendar with another person, or a file shared. So I think we can stick with showing just the display name :) We can always show the full name, which would help to differentiate to a certain extent.

pulsejet:
I'm not talking about UX. I think the UX is fine, but there is a correctness problem that stems from the lexical DAV API (i.e. we can no longer uniquely identify the two albums).

If we just (for example) used IDs to identify albums, then this issue would never arise. It would then even be perfectly valid to have two albums with the same name for one user, which we can't have right now because of the artificial limitation that the DAV API adds.

artonge:
Revert and improvement of the UI here: https://github.com/nextcloud/photos/pull/1786

Issue #1396, opened 2022-10-19
Uploading files to an album does not work


nicokaiser:
**Describe the bug**
Uploading files to an album does not seem to work. 

**To Reproduce**
1. Navigate to Albums
2. Click on "+ New album"
3. Set the Name of the album
4. Click on "Create album"
5. Click "+ Add"
6. In the Popup, click "+ Add"
7. Select files to add to the album

Photos are uploaded to the /Photos folder, but the album stays empty. The browser console does not show any error messages.

**Expected behavior**
I'd expect the uploaded photos to be added to the album.

**Desktop (please complete the following information):**
 - OS: macOS
 - Browser: Chrome

**Server:**
Nextcloud 25.0.0

skjnldsv:
I will need a screenshot of your network requests.
Before you do anything, still on the development tools, there is a tab called `network`. Click on it and then the xhr filter. Do your action and screenshot the network log like this:
![capture d ecran_2018-09-27_21-49-46](https://user-images.githubusercontent.com/14975046/46170857-43548000-c29f-11e8-9403-db72713dd9c1.png)

nicokaiser:
![Bildschirmfoto 2022-10-19 um 19 35 09](https://user-images.githubusercontent.com/238631/196764197-c05ae356-2b71-41a5-be7e-6c78557411d9.png)

Here is a screenshot. The upload of the two jpg files worked, but they do not appear in the album.

skjnldsv:
Yep, seems like there is a request missing.
The files gets uploaded in your Photos folder though, right?

nicokaiser:
> The files gets uploaded in your Photos folder though, right?

Yes, correct.

nicokaiser:
(I only found this while looking for a convenient way to add many photos to an album. Since there is currently no way to select multiple photos, e.g. by Shift-clicking the last photo, I tried to upload them to an album directly... I'll add a separate feature-request for the multi-selection feature)

Gordin508:
I also noticed this using Nextcloud version 25.0.2.

The problem is that the "Add photos to this album" dialog does not seem to get updated with new images uploaded through its "Add" option. When I reopen the dialog once, the newly added image is shown by name but the thumbnail is missing (error 500 for the `?x=64&y=64` thumbnail). Reopening it twice shows the newly added image. Switching to another page and back again also shows the image in the dialog.

[...]

PR #1782, opened 2023-05-10
Reset file list on upload in FilesPicker

artonge:
Fix https://github.com/nextcloud/photos/issues/1396

[Screencast from 2023-05-11 12-06-59.webm](https://github.com/nextcloud/photos/assets/6653109/1d1162f3-12c1-488f-a38a-00328e18d082)


skjnldsv:
Does that means it will always flash end re-render the entire list? Could we find a middle ground and just append the new files?

skjnldsv (approved):

PR #1786, opened 2023-05-11
Revert using display name in shared albums' name

artonge:
Revert part of https://github.com/nextcloud/photos/pull/1781 to keep unique album names, and improve rendering of cover and header of shared albums

Based on concerns expressed here: https://github.com/nextcloud/photos/pull/1781#issuecomment-1542599483

Before | After
-- | --
Screenshot from 2023-05-11 14-39-46 | Screenshot from 2023-05-11 14-34-08
Screenshot from 2023-05-11 14-39-32 | Screenshot from 2023-05-11 14-34-15


nimishavijay:
This looks great! Since this PR purpose is to make it clear who shared the item with you, I would suggest showing the full name along with the avatar, using a user bubble (without the primary color), screenshots attached. What do you think?

nimishavijay (approved):
🚀 :)

pulsejet (approved):
Love that bubble 💯

jancborchardt (commented):
Awesome, looks much better with the userbubble.

Detail: In the grid list, the "Shared by" could go on the same line as date and "X items" as that's not so long @artonge? Then there is less of a vertical change between shared and non-shared albums.

artonge:
> Detail: In the grid list, the "Shared by" could go on the same line as date and "X items" as that's not so long @artonge? Then there is less of a vertical change between shared and non-shared albums.

I am worried that this would overflow in most cases, so keeping it as is.

Issue #1322, opened 2022-10-10
Automate testing of photos 2.0 features


artonge:
Here is a list of feature that photos 2.0 should support, but does not have automated testing yet:

- [ ] Alice share album with Bob and Charlie, both add photos, all three should see each other photos
- [ ] Same with a group.
- [x] Download all shared album files from content view
- [x] Download individual shared album file from content view
- [x] Download individual shared album file from viewer
- [x] Delete share from recipient account
- [ ] Add file to shared album from timeline (list shared albums in album picker)
- [x] Two shared albums with the same name, we should prefix with userid
- [ ] Deleting a file added to an album should not break the album
- [ ] Removing a collaborator should remove its photos
- [ ] Deleting a user should delete its albums
- [ ] Deleting a user should remove it from the collaborators list
- [ ] Prevent user to delete its received shared when it is via a group
- [ ] Make code resistant if for file added from a deleted share
- [x] Removing a file from an album just after adding it does not work as the filename does not contains the fileid
- [x] Favorite and download a file in a classic album
- [x] Favorite and download a file in a shared album
- [x] Favorite and download a file in a timeline
- [ ] Download files in albums

blizzz:
moved to 26, feel free to put it back to a 25 maintenance milestone if that was intended.

PR #1826, opened 2023-05-24
Add more e2e tests

artonge:
Fix https://github.com/nextcloud/photos/issues/1322

## Added tests

- Test creating, accessing, and deleting public links
- Test adding files to a shared album from timeline
- Test multiple collaborators for shared albums
    - See each other pictures
    - Removing a collaborator removes its pictures 
    - Can remove each other pictures
- Testing that deleting a user removes it from the collaborators list if albums
- Testing that deleting a user deletes its albums
- Test if deleting file break an album

## Enhancements

- Use `NcSelect` for collaborator selection form
- Support browsers that do not support caches
- Add jest's types for tests files
- Improve resilience of 2e2 tests

## Fixes

- Prevent deletion of a received album when it is received from group share
- List received shared albums in album picker
- Remove user from collaborators when it is deleted

### DIFF
--- a/.github/workflows/cypress.yml
+++ b/.github/workflows/cypress.yml
@@ -17,20 +17,20 @@ jobs:
 
     steps:
       - name: Checkout app
-        uses: actions/checkout@v3
+        uses: actions/checkout@8e5e7e5ab8b370d6c329ec480221332ada57f0ab # v3.5.2
 
       - name: Install server dependencies
         run: composer install
 
       - name: Read package.json node and npm engines version
-        uses: skjnldsv/read-package-engines-version-actions@v2.0
+        uses: skjnldsv/read-package-engines-version-actions@0ce2ed60f6df073a62a77c0a4958dd0fc68e32e7 # v2.1
         id: versions
         with:
           fallbackNode: "^14"
           fallbackNpm: "^7"
 
       - name: Set up node ${{ steps.versions.outputs.nodeVersion }}
-        uses: actions/setup-node@v3
+        uses: actions/setup-node@64ed1c7eab4cce3362f8c340dee64e5eaeef8f7c # v3.6.0
         with:
           cache: "npm"
           node-version: ${{ steps.versions.outputs.nodeVersion }}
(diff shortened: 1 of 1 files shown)


artonge:
Screenshot from 2023-05-24 14-53-41

@Pytal I remember you add the same issue some time ago, is there a fix?
Ideally, I would like the input not to be focused when the pop-up appears.

Edit:
I am using this hack for now. Better idea welcome :)
```js
:dropdown-should-open="({open, search}) => open && search !== ''"
```

Pytal:
🧐 was fixed in https://github.com/nextcloud-deps/vue-select/pull/14

Before the fix this happened because the dropdown itself would overflow the container but in this case it doesn't look like it would overflow, perhaps the fix above though may provide a hint to the bug in this specific case @artonge?

Pytal:
> Edit: I am using this hack for now. Better idea welcome :)
> 
> ```js
> :dropdown-should-open="({open, search}) => open && search !== ''"
> ```

If the dropdown never needs to be placed above the input then `:append-to-body="false"` should work @artonge

AndyScherzinger (approved):

cypress.config.ts
```diff
@@ -17,9 +17,11 @@ export default defineConfig({
 	viewportWidth: 1280,
 	viewportHeight: 720,
 
+	requestTimeout: 20000,
+
 	// Tries again 2 more times on failure
```
Pytal, line 22:
```suggestion
```
```
runMode: 0,
```

Pytal (approved):

skjnldsv:
You're close to success it seems!!

artonge:
/backport to stable27 1d8058a50c4c78f635e6c4d511f6e7bbc28c7233

artonge:
/backport 1d8058a50c4c78f635e6c4d511f6e7bbc28c7233to stable27

artonge:
/backport 1d8058a50c4c78f635e6c4d511f6e7bbc28c7233to to stable27

Issue #1833, opened 2023-05-28
Photos won't show up in Photos app (through browser) when image doesn't have a specific metadata 


adriano-pinaffo:
[**Describe](url) the bug**
After uploading some image files the Photo app stops showing the images (thumbnails don't load).
===> Issue identified (check below)
===> Fix suggested (check below)

**To Reproduce**
1. Upload specific images into the Photo folder (or any folder that is in the scope of the Photo app). A photo example is provided here.
2. Open the Photo app (no thumbnail will show up)

**Expected behavior**
Thumbnails show up in Photo app

**Screenshots**
![nextcloud-photo-issue](https://github.com/nextcloud/photos/assets/28174227/2769dc19-c3cf-449b-bc63-4dd511ce51ab)

**Desktop (please complete the following information):**
 - OS: Linux Mint
 - Browser: Brave Version 1.46.153 Chromium: 108.0.5359.128, also happens in Firefox
 - Version: Nextcloud 26 (AIO v5.1.0)

**Issue identified**
The problem happens only in the browser in the front-end of the Photos app. When generating the thumbnails. the object `file` has a property `fileMetadataSize` which is a string representing the width and height of the image. The program will parse it to extract the width and height as properties of `fileMetadataSizeParsed`. The problem happens when the property `fileMetadataSize` is an empty string, this throws an error because `JSON.parse()` can't parse empty strings and the Nullish coalescing operator (`??`) won't assign `'{}'` because `.replace()` won't return `null` and the Optional chaining (`?.`) won't break the chain since an empty string is not null.
![image](https://github.com/nextcloud/photos/assets/28174227/aba50a65-20e2-4a54-8f7b-3ee73a53e1cd)

Note line 54 in the snippet below (in `photos/src/store/files.js`):
```
 52       if (file.fileid >= 0) {
 53         if (file.fileMetadataSize?.length > 1) {
 54           file.fileMetadataSizeParsed = JSON.parse(file.fileMetadataSize?.replace(/&quot;/g, '"') ?? '{}')
 55           file.fileMetadataSizeParsed.width = file.fileMetadataSizeParsed?.width ?? 256
 56           file.fileMetadataSizeParsed.height = file.fileMetadataSizeParsed?.height ?? 256
 57         } else {
 58           file.fileMetadataSizeParsed = { width: 256, height: 256 }
 59         }
 60       }
```
Line 54 causes the following error:
```
[...]

PR #1834, opened 2023-05-28
Prevent error on generating thumbnails

adriano-pinaffo:
When generating the thumbnails the parsing throws an error if the file property representing the image size metadata is an empty string. This protects the string by making it be a `"{}"` if the string is empty. This is described in [#1833](https://github.com/nextcloud/photos/issues/1833).

### DIFF
--- a/src/store/files.js
+++ b/src/store/files.js
@@ -51,6 +51,7 @@ const mutations = {
 
 			if (file.fileid >= 0) {
 				if (file.fileMetadataSize?.length > 1) {
+					file.fileMetadataSize = file.fileMetadataSize === '' ? '{}' : file.fileMetadataSize
 					file.fileMetadataSizeParsed = JSON.parse(file.fileMetadataSize?.replace(/&quot;/g, '"') ?? '{}')
 					file.fileMetadataSizeParsed.width = file.fileMetadataSizeParsed?.width ?? 256
 					file.fileMetadataSizeParsed.height = file.fileMetadataSizeParsed?.height ?? 256


marcelklehr:
I've recently fixed a bug in recognize that would cause this, so this might not be necessary, but we can add it for good measure.

artonge:
@marcelklehr can you point to the fix and @adriano-pinaffo can you check if this fix the issue for you?
I would rather fix the issue than silent the error.

marcelklehr:
here you go: https://github.com/nextcloud/recognize/commit/0fdb558f46df1141d4f205d3eb6fa09104a73c71

marcelklehr:
@adriano-pinaffo Could you disable recognize to check whether this bug stems from recognize then?

adriano-pinaffo:
I disabled Recognize, went back to Photos and CTRL+F5 it... still, Images are not shown

image

SeanDS:
@marcelklehr I had this issue with Photos before installing Memories. Since installing Memories I am able to view the photos there but not in the Photos app.

It might be useful to note, that I only started to get this issue in Photos after I ran `files:scan`, `files:repair-tree`, `files:cleanup` and `files:scan-app-data` operations with `occ` after I manually modified the file system. This is probably related.

SeanDS:
Update: I fixed it by running `occ files:scan --all` then `occ files:scan --all --generate-metadata`, also having updated to 26.0.1 to 26.0.5.

AndyScherzinger:
@artonge what do you think still needed or fixed and can be closed?

artonge:
The code has changed since then. I don't see how the error could occur at the time, and don't see how it could occur now. Feel free to reopen if needed.

PR #1845, opened 2023-06-05
Avoid unnecessary API requests 

Dymirt:
In the fetchFaces method, we can remove the check if (Object.keys(this.faces).length) before making the API request. Since the fetchFaces method is called in the beforeMount hook, it will already be executed only once during the component's lifecycle. Therefore, we don't need to check if the faces data is already present in the Vuex store before making the API request.


Dymirt:
By combining the operations into a single loop, we reduce the number of iterations over the array and minimize unnecessary processing, leading to improved performance.

artonge:
Also, it does not look like we are doing less API requests.

Dymirt:
> Also, it does not look like we are doing less API requests.

You're right. there is same amount of an API requests. But using loop is more readable in first, additionally, consolidating the operations into a single loop can improve memory usage by avoiding the creation of intermediate arrays. This can be beneficial in scenarios where memory consumption is a concern.

This is my first commit in to nextcloud, so im trying to make some small changes. On my nextcloud I have more than 100 people recognised and some of them have more than 1000 photos. So im trying to make les requests, and more efficient loading sice there is too many requests and all photos per person is loading on main Persons page.
My logs looks like:
[Debug] [DEBUG] photos: [FetchFacesMixin] Fetched 1235 new files: – {0: "99030", 1: "99159", 2: "99160", …} (photos-main.js, line 2)

src/mixins/FetchFacesMixin.js
```diff
@@ -68,9 +68,6 @@ export default {
 				return
 			}
 
-			if (Object.keys(this.faces).length) {
-				return
-			}
```
marcelklehr, line 73: I'd vote to keep this just for sanity

src/mixins/FetchFacesMixin.js
```diff
@@ -120,14 +117,19 @@ export default {
...
+				for (const file of fetchedFiles) {
+				  const processedFile = genFileInfo(file);
+				  processedFile.filename = he.decode(processedFile.realpath).replace(`/${getCurrentUser().uid}/files`, `/files/${getCurrentUser().uid}`);
+				  processedFile.faceDetections = JSON.parse(he.decode(processedFile.faceDetections));
+			
+				  processedFiles.push(processedFile);
+				  fileIds.push(String(processedFile.fileid));
+				}
+			
+				this.appendFiles(processedFiles);
```
marcelklehr, line 132: I think it's a nice change. Not sure if strictly necessary but surely improves readability a bit I think.

artonge:
> all photos per person is loading on main Persons page.

Indeed, any reason we are doing this @marcelklehr?

marcelklehr:
Thank you for looking into making the faces views more performant @Dymirt :blue_heart: 

I think the real gold nugget is this line here: https://github.com/nextcloud/photos/blob/master/src/components/FaceCover.vue#L115

We should try to only load the files per face when that face is actually opened. the face directories should already have a nbItems property that contains the number of faces, but it doesn't have any properties for the cover images. I guess it would be best if recognize supplies that. I'll come up with a way to add that, then we can implement this. Feel free to hop by in https://matrix.to/#/#marcelklehr_recognize:gitter.im to coordinate on this if you like :) :wave:

marcelklehr:
> Indeed, any reason we are doing this @marcelklehr?

lack of foresight :D

src/mixins/FetchFacesMixin.js
Dymirt: I've added this "if" back.

Dymirt:
@marcelklehr Can you please do code review?

src/mixins/FetchFacesMixin.js
Dymirt: Thank you!

marcelklehr:
It seems @artonge has been refactoring the photos frontend, so we could now reuse his collections interfaces for faces.

marcelklehr:
@Dymirt DCO  is still missing. Could you sign your commits?

artonge:
> It seems @artonge has been refactoring the photos frontend, so we could now reuse his collections interfaces for faces.

More, is coming with https://github.com/nextcloud/photos/pull/1621, if you can wait 1 week :D

Dymirt:
> @Dymirt DCO is still missing. Could you sign your commits?



> @Dymirt DCO is still missing. Could you sign your commits?

I have trouble with sign my previous commits.

marcelklehr:
> More, is coming with https://github.com/nextcloud/photos/pull/1621 for stores, should be merged in a week or so.

Alright, then let's get this PR in before that and wait for @artonge 's PR :)

PR #1851, opened 2023-06-06
[stable25] Fix updating file from albums

artonge:
Probably a bad copy-past, `$this->userfolder` and `$this->file` do not exist.


skjnldsv:
Good catch!

PR #1854, opened 2023-06-06
[stable24] Bump @nc/vue and @nc/axios to fix 'npm audit'

artonge:


artonge:
Some dependencies needed to be bumped, and I understand that `npm audit fix` do not handle those cases.

szaimen:
> Some dependencies needed to be bumped, and I understand that `npm audit fix` do not handle those cases.

So this PR is not only `npm run fix` ?

artonge:
> So this PR is not only `npm run fix` ?

Nope, it's a bit more.

PR #1856, opened 2023-06-06
Update view when layout setting changes

artonge:
[Screencast from 2023-06-06 17-34-44.webm](https://github.com/nextcloud/photos/assets/6653109/822fb253-3812-4851-96d7-367640e3a2c9)


src/components/File.vue
```diff
@@ -98,7 +97,6 @@ export default {
 		Star,
 		VideoIcon,
 	},
-	mixins: [UserConfig],
```
artonge, line 101: Seems unused.

Pytal (approved):

PR #1860, opened 2023-06-07
Fix/avoid unnecessary api requests

Dymirt:


marcelklehr:
@artonge Not sure why the compile command fails

artonge (commented):
I am still unsure about the need for this change.
Didn't we say that the issue was with loading the content of all the faces?
This is just a rewrite of the code without any benefit, no?

artonge:
Alright. Then you'll need to add the result of `npm run build` to your commit. :)

marcelklehr:
@Dymirt I've started to work on the actual performance fix, since I realized it would probably have been a bit too complex for you, as it also requires changes in recognize. Here we go: https://github.com/nextcloud/photos/pull/1881

PR #1882, opened 2023-06-13
Split sections in independent lists

artonge:
Quite a big change as I introduced the concept of section in `TiledLayout` and `VirtualScrolling`.

Sections are now the top structure that those components receive. Each section then contain a list of items for `TiledLayout` and a list of rows for `VirtualScrolling`.

The PR:

- Further enhances the accessibility of our list.
- Fix the disappearance of headers, as they were previously removed as any other row.
- Simplify the logic of the tiled layout, as it is not concerned with headers now.

Notice:

- The headers slightly differ in the screenshots as I also switch from `h3` to `div` as `h3` is a bit arbitrary, why not h2? And does not make sense in places like `FilePicker`.

Before | After
-- | --
Screenshot from 2023-06-21 17-09-55 | Screenshot from 2023-06-21 17-11-38
Screenshot from 2023-06-21 17-07-17 | Screenshot from 2023-06-21 16-58-40

Before: 
```
div
  ul // Unique list
    li // Section header #1
      h3
    li // Item
      img
    li
      img
    li // Section header #2
      h3
    li
      img
```

After
```
div
  div // Section container #1
    h2 // Section header
    ul // Section list
      li // Item
        img
      li
        img
  div // Section container #2
    h2
    ul
      li
        img
```

Fix https://github.com/nextcloud/photos/issues/1665


JuliaKirschenheuter:
Thank you @artonge! Looks very good!

Regarding headers: i would like to stay with `<h3>` because semantically they are headers. And semantic in our case is really important. We can just replace `<div>` with `<h3>` for `section-header` and remove `margin: 12px 0;` for this headers. What do you think?

artonge:
@JuliaKirschenheuter thanks for the review :)

My concern was that h3 does not make sense in every cases, like in the file picker for exemple. Even for the timelines, I am not sure whether h3 is fine or if we should use h2.

What I can do is to allow the different views to specify which kind of element they want for headers, this way, timelines and file picker can have something different.

But which level of header should they use in both cases ?

src/views/Timeline.vue
```diff
@@ -89,13 +89,13 @@
 			:base-height="isMobile ? 120 : 200"
 			:empty-message="t('photos', 'No photos or videos in here')"
 			@need-content="getContent">
-			<template slot-scope="{file, distance}">
-				<h3 v-if="file.sectionHeader"
+			<template slot-scope="{file, isHeader, distance}">
+				<div v-if="isHeader"
```
JuliaKirschenheuter, line 93:
```suggestion
				component :is="isHeader ? h3 : div"
```

src/components/FilesPicker.vue
```diff
@@ -43,13 +43,13 @@
 				:section-header-height="50"
 				:scroll-to-section="targetMonth"
 				@need-content="getFiles">
-				<template slot-scope="{file, height, distance}">
-					<h3 v-if="file.sectionHeader"
+				<template slot-scope="{file, height, isHeader, distance}">
+					<div v-if="isHeader"
```
JuliaKirschenheuter, line 47:
```suggestion
					component :is="isHeader ? h3 : div"
```

JuliaKirschenheuter (approved):
quite a big change. but so far looks good to me, tested.

JuliaKirschenheuter:
> My concern was that h3 does not make sense in every cases, like in the file picker for example.

I would suggest to use `<h2>` for the navigation on the left side and `<h3>` for the titles inside of the content. That seems logically to me.

JuliaKirschenheuter:
> What I can do is to allow the different views to specify which kind of element they want for headers, this way, timelines and file picker can have something different.

> allow the different views to specify which kind of element they want for headers

Could you paste some screenshots for different scenarios please?

JuliaKirschenheuter:
> If a modal dialog is coded correctly, the entire webpage disappears and leaves only a window with a keyboard trap.
https://medium.com/@web-accessibility-education/which-heading-level-should-dialog-modals-have-7b3df89437f0

(that means that everything behind the modal window doesn't play any role)

Examples from https://www.w3.org/WAI/ARIA/apg/patterns/dialog-modal/examples/datepicker-dialog/ and https://www.w3.org/WAI/ARIA/apg/patterns/alertdialog/examples/alertdialog/ showing that modal dialogs starting with `<h2>`. I think it is ok to use `<h2>` for the titles in file picker.

Please don't forget to give a description (`aria-label`) for the modal dialog, something like "Add files to your album".

PR #1895, opened 2023-06-19
Add unselect all button

artonge:
Add "Unselect all" button.

I went the easy way, as dynamically displaying action from the three dot menu based on the available space would require some changes in `@nc/vue`

I also took the opportunity to prevent selection in Places and Public album views as no actions can be done from those.

View | After
-- | --
Timeline | Screenshot from 2023-07-05 14-41-31
Albums | Screenshot from 2023-07-05 14-41-41


Fenn-CS (approved):
LGTM, thanks!

Issue #1674, opened 2023-03-02
[BITV] 9.2.5.3/4.1 - The visible label "Add" does not appear in the accessible name of the button in the same form as it is overridden by the use of aria-label="Actions". In this case, it cannot be guaranteed that the button will be activated through voice input. (1)


AndyScherzinger:
- ~https://github.com/nextcloud/photos/pull/2013~
- [x] https://github.com/nextcloud-libraries/nextcloud-vue/pull/4447
- [x] https://github.com/nextcloud/photos/pull/2020
- [x] Check

![0e4660212b80e42b753b1adcc08a8370](https://user-images.githubusercontent.com/1315170/222545383-0d95b7b2-b2b1-4ef8-8884-b957cfdc23fc.png)

<details>
<summary>Details</summary>

https://report.bitvtest.de/default-en/d63601ac-cb34-4645-8256-66bec78964a0.html#checkpoint-c6347c127a-v4-n1

</details>

Pytal:
Conflicting `aria-label` no longer present
![image](https://github.com/nextcloud/photos/assets/24800714/bfb12ed8-f824-4481-8173-2878a9397067)

PR #2013, opened 2023-08-19
fix(a11y): Activate add button with voice input

Pytal:
- For https://github.com/nextcloud/photos/issues/1674

Ensure users who use voice input can activate the add button

Ref: https://ashleemboyer.com/blog/don-t-overwrite-visual-labels-with-aria-label#how-do-i-fix-mismatched-aria-labels-and-visual-labels


ShGKme (approved):
Should we remove using default value for `ariaLabel` in `NcActions` when `menuTitle` is present?

Pytal:
> Should we remove using default value for `ariaLabel` in `NcActions` when `menuTitle` is present?

Since v8 is still in beta we can break things :D, so let's go with that!

Pytal:
PR at https://github.com/nextcloud-libraries/nextcloud-vue/pull/4447

Pytal:
PR merged

Superseded by bump

PR #2052, opened 2023-09-19
Recycle DOM elements in virtual scrolling

artonge:


skjnldsv (approved):
Why so much bundle changes :sweat_smile:

PR #2087, opened 2023-10-12
Fix buttons on new album modal window

JuliaKirschenheuter:
* Part of https://github.com/nextcloud/server/issues/36965
* New album modal window

## Summary

Fix fit of buttons on new album modal window

🏚️ Before | 🏡 After
---|---
image | Screenshot from 2023-10-17 15-54-36


src/components/Albums/AlbumForm.vue
```diff
@@ -292,9 +292,6 @@ export default {
 	}
 
 	.form-buttons {
-		display: flex;
-		justify-content: space-between;
-
```
artonge, line 297: Or `flex-direction: column` ?

artonge (approved):

src/components/Albums/AlbumForm.vue
```diff
@@ -294,6 +294,7 @@ export default {
 	.form-buttons {
 		display: flex;
 		justify-content: space-between;
+		flex-direction: column;
```
szaimen, line 297: always? I thought only on mobile?

JuliaKirschenheuter: Screenshot from 2023-10-13 11-18-04

desktop

szaimen: Or mybe just use flex-wrap? Wdyt?

szaimen:
```suggestion
		flex-wrap: wrap;
```

JuliaKirschenheuter: @szaimen nope ;(
Screenshot from 2023-10-16 15-56-37

szaimen: Hm... Is there some reason why that does not work? Should usually be shown in the inspector tools...

JuliaKirschenheuter: does it works for you? For me: styles are there in dev tools.

src/components/Albums/AlbumForm.vue
```diff
@@ -312,4 +313,12 @@ export default {
 .left-buttons {
 	flex-grow: 1;
 }
+
+@media (max-width: 1280px) {
```
ShGKme, line 317: To be a bit more consistent with other places in Photos and 1024 breakpoint.

```suggestion
@media only screen and (max-width: 1020px)
```

szaimen:
```suggestion
@media (max-width: 1024px) {
```

PR #2098, opened 2023-10-21
fix: open settings button

ShGKme:
A part of https://github.com/nextcloud/server/issues/37092

> Error: Element [`li`](https://html.spec.whatwg.org/multipage/#the-li-element) not allowed as child of element [`div`](https://html.spec.whatwg.org/multipage/#the-div-element) in this context. (Suppressing further errors from this subtree.)

`<NcAppNavigationItem>` - renders as a list item `<li>` and is not allowed in the navigation footer without UL wrapper.

But this part of the UI is not a list or navigation link, it only opens a modal.

This PR replaces **navigation item** with **button**.

The position of the icon is still a bit not in line, but IMO it should be fixed on the Nextcloud/vue side.

Also fixed new Vue 2.7 `:deep` selector usage.

Before | After
---|---
image | image


susnux (approved):
We probably should have this in the library as an alternative to [NcAppNavigationSettings](https://nextcloud-vue-components.netlify.app/#/Components/App%20containers/NcAppNavigation?id=ncappnavigationsettings)

JuliaKirschenheuter (approved):

ShGKme:
> We probably should have this in the library as an alternative to [NcAppNavigationSettings](https://nextcloud-vue-components.netlify.app/#/Components/App%20containers/NcAppNavigation?id=ncappnavigationsettings)

Just a `boolean` prop to show the settings button? Hmm, makes sense.

PR #2099, opened 2023-10-21
Fix new album form layout

ShGKme:
A part of https://github.com/nextcloud/server/issues/37092

> Error: The element `label` must not appear as a descendant of the `label` element.

- Remove incorrect `<label>` wrapper
- Use `<NcTextField>'s` icon feature instead of separated icon
- Use recommended 20px icon size on all icons
- Use CSS variables for spacing
- Remove right margin from buttons
- Remove `aria-label`-s. Buttons with texts should not have aria-label that is different from the button text. 

Before | After
---|---
image | image


susnux (approved):
clean and looks much better!

susnux:
Just fix the node CI :)

ShGKme:
> Just fix the node CI :) 

Would you believe if I say, I did npm ci, npm run build, got status is clear and node is 20 with npm 9 🙈

JuliaKirschenheuter (approved):

ShGKme:
Just rebased and re-compiled

PR #2103, opened 2023-10-25
Filter out some permission for public albums

artonge:
Remove some permissions for photos in public albums.

This prevent the viewer to display some unavailable actions, like delete and edit.

The G permissions was previously removed but no idea why. I replaced it with the R permissions as this one is useless in Photos.

Ref: https://docs.nextcloud.com/server/latest/developer_manual/client_apis/WebDAV/basic.html#webdav-basics


skjnldsv (approved):
:elephant:

PR #2104, opened 2023-10-25
Use new metadata API for providers

artonge:
Need https://github.com/nextcloud/server/pull/40761
Part of https://github.com/nextcloud/server/issues/40676


lib/MetadataProvider/ExifMetadataProvider.php
```diff
@@ -0,0 +1,136 @@
...
+		}
+
+		$node = $event->getNode();
+
+		if (!$node instanceof File) {
+			return;
+		}
+
+		$path = $node->getPath();
+		if (str_starts_with($path, 'appdata_') || str_starts_with($path, 'files_versions/') || str_starts_with($path, 'files_trashbin/')) {
```
artonge, line 56: Check if it is triggered when you move a file to the trashbin.
Same with `files_versions`
Check that `$path` is appropriate to check.
Could be put in File API.

artonge: Quick tests show that `NodeWrittenEvent` is not triggered for versions, trash, and appdata items.

artonge: So this can probably be safely removed

lib/MetadataProvider/SizeMetadataProvider.php
```diff
@@ -0,0 +1,72 @@
...
+		}
+
+		$node = $event->getNode();
+
+		if (!$node instanceof File) {
+			return;
+		}
+
+		$path = $node->getPath();
+		if (str_starts_with($path, 'appdata_') || str_starts_with($path, 'files_versions/') || str_starts_with($path, 'files_trashbin/')) {
```
skjnldsv, line 53: What about the other way around and only check for exact `files/` nodes?

artonge: Group folders and external folders? But their paths might be within `/files`. 
I am actually not sure whether the check is needed, see above comment

artonge: It depends on the kind of event that is listened by the Metadata API

lib/Sabre/PropFindPlugin.php
```diff
@@ -111,21 +105,8 @@ public function propFind(PropFind $propFind, INode $node): void {
...
-					if ($node->getFile()->hasMetadata('size')) {
-						$sizeMetadata = $node->getFile()->getMetadata('size');
-					} else {
-						$sizeMetadata = $this->metadataManager->fetchMetadataFor('size', [$node->getFile()->getFileId()])[$node->getFile()->getFileId()];
-					}
-
-					return $sizeMetadata->getValue();
-				});
+			foreach ($node->getFileInfo()->getMetadata() as $metadataKey => $metadataValue) {
+				$propFind->handle(FilesPlugin::FILE_METADATA_PREFIX.$metadataKey, $metadataValue['value']);
```
artonge, line 109: This has changed

lib/Service/MediaPlaceManager.php
```diff
@@ -46,31 +47,33 @@ public function setPlaceForFile(int $fileId): void {
 		$this->placeMapper->setPlaceForFile($place, $fileId);
 	}
 
-	public function updatePlaceForFile(int $fileId): void {
-		$place = $this->getPlaceForFile($fileId);
-
-		if ($place === null) {
-			return;
-		}
+	// public function updatePlaceForFile(int $fileId): void {
```
artonge, line 50: To delete

ArtificialOwl (approved):
🐘

lib/DB/Place/PlaceMapper.php
artonge, line 1: - Make more use of `IFilesMetadataManager`
- Delete methods that are not useful anymore

ArtificialOwl (approved):
back-end is 👍

PR #2128, opened 2023-11-14
metadata: fix crash when datetime has invalid format

pulsejet:
For any reason if the date time is not in the correct format, the entire scan process crashed.


artonge (approved):

artonge:
Thanks for testing and for the correction :heart: 
Did you have any issue for the second fix? As the date format is supposed to match the regexp, I would think that there couldn't be any error. In any case, let's keep it :)

pulsejet:
> As the date format is supposed to match the regexp, I would think that there couldn't be any error.

There _might_ be. For example `PXL_99999999_999999.JPG` would match the regex but I don't know if PHP can parse this correctly.

PR #2135, opened 2023-11-16
Fix EXIF display conditions

artonge:
Those are more specific and prevent errors.


emoral435 (approved):
Seems good to me, the new conditions look good!

PR #2154, opened 2023-11-24
add setting 'disable_place'

ArtificialOwl:


artonge (approved):
Can you make it plural? placeS

artonge:
Same with method name and such

ArtificialOwl:
@artonge please re-review and merge if fine

lib/Command/UpdateReverseGeocodingFilesCommand.php
```diff
@@ -49,6 +49,10 @@ protected function configure(): void {
 	 */
 	protected function execute(InputInterface $input, OutputInterface $output): int {
 		try {
+			if (!$this->rgcService->arePlacesEnabled()) {
+				throw new \Exception('Place is disabled');
```
artonge, line 53:
```suggestion
				throw new \Exception('Places is disabled');
```

ArtificialOwl: are disabled ? :]

PR #2182, opened 2023-12-06
Fix checkbox layout

artonge:
Before | After
-- | --
image | image
image  | image

Issue due to an internal change in `@nc/vue`


artonge:
/backport c7fd86d9936e0e754678e4b459d7908e16312ff2 to stable27

src/components/File.vue
```diff
@@ -335,14 +334,14 @@ export default {
 		width: fit-content;
 
 		// Make the checkbox background round on hover.
```
emoral435, line 336: Right here, a nitpick is that now we want the background to always be present for accessibility reasons, no? So the comment could remove the hover part

PR #2205, opened 2023-12-18
Fix Fatal Error in Album setCollaborator function - redefining computeKey function

frederikb96:
When calling setCollaborator function two times after another, computeKey is redefined and leads to a fatal error. Fixed by defining it as an anonymous function instead :)

I wrote an app, which calls this function, and thus I discovered that problem.

### DIFF
--- a/lib/Album/AlbumMapper.php
+++ b/lib/Album/AlbumMapper.php
@@ -440,12 +440,12 @@ class AlbumMapper {
 		$existingCollaborators = $this->getCollaborators($albumId);
 
 		// Different behavior if type is link to prevent creating multiple link.
-		function computeKey($c) {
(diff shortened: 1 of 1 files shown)


artonge (approved):

artonge:
Thanks @frederikb96, can you fix the DCO?

frederikb96:
Thanks, I will look into it tomorrow 🙂

frederikb96:
@artonge Ok, I guess that's it right (I am new to this :)), thanks.

frederikb96:
> @frederikb96 Php cs is also red. Running `php run cs:fix` should be enough. Cypress tests failure seem unrelated. :)

@artonge Ok php cs is fixed now :)

artonge:
Thanks a lot @frederikb96 !

PR #2212, opened 2023-12-22
fix: glitch buttons on mobile

ShGKme:
* Fixes https://github.com/nextcloud/server/issues/41852

Buggy reaction on slot default content update in Vue 2.

Without `#default` there is always a default slot with probably empty content, that is always passed as a content. With `template v-if` it is passed as a render function and not passed at all when there is no content to be set.

Before | After
---|---
photos-glitch-before | photos-glitch-after


ShGKme:
I cannot understand why my local build doesn't work for Photos CI 🙈

JuliaKirschenheuter (approved):
Thank you!

Issue #2237, opened 2024-01-08
[BITV]: Contrast of a title is not enough 


JuliaKirschenheuter:
On index.php/apps/photos/folders

![Screenshot from 2024-01-08 17-58-40](https://github.com/nextcloud/photos/assets/6078378/783b1d7d-3a4a-4762-b513-2135a601d72c)

PR #2255, opened 2024-01-17
enh(a11y): Update folder design to fix contrast issues

Pytal:
- Fix https://github.com/nextcloud/photos/issues/2237

### Summary

- Update folder component to match album component design
- Make page mobile-friendly at 320px
- Test with NVDA

Before | After
--- | ---
image | image
image | image

| Albums |
| --- |
| image |


Pytal:
Not sure about design, any suggestions @szaimen @nimishavijay?

nimishavijay:
Hmm, not too sure about this either. What do you think about making it similar to the design of albums? Some changes will be needed to differentiate it from albums, such as a smaller thumbnail and a folder icon before the title. What do you think?

Pytal:
> Some changes will be needed to differentiate it from albums, such as a smaller thumbnail and a folder icon before the title. What do you think?

Made it the same size as albums with 350x350px thumbnails to be consistent, but open to a smaller size suggestion if you'd think it'd be better design-wise @nimishavijay?

nimishavijay:
Nice, already looks better! Not a blocker, but I  would indeed suggest a smaller size, so that the focus is more on the name of the folder. How about 200px? :)

artonge (approved):
How does it look when folder and images are mixed together?

Pytal:
> How does it look when folder and images are mixed together?

Less noticeable now with 200px thumbnails but basically just scales it up slightly from the file-only grid

image

artonge (approved):

JuliaKirschenheuter (approved):
Works and looks great 😊

Issue #2238, opened 2024-01-08
[BITV]: Not possible to add already selected photos to the album via keyboard


JuliaKirschenheuter:
1. Create an album
2. Try to select some photos and add them to the album via keyboard

![Peek 2024-01-08 18-05](https://github.com/nextcloud/photos/assets/6078378/e349391f-40c2-4e1c-9e75-2735872fdde7)

It is not possible to reach "Add to ..." and "Add" buttons via keyboard

Pytal:
This issue seems to involve multiple dependencies including @nextcloud/files, @nextcloud/upload, and @nextcloud/vue with changed beta APIs making it slightly confusing

When I checkout the @nextcloud/upload package to v1.0.4 and link it locally + src changes the bug is resolved but updating to v1.0.4 from npm in package.json does not seem to work, any ideas why this might be @ShGKme?

Pytal:
Draft https://github.com/nextcloud/photos/pull/2256 was closed due to scope creeping into Folder/Node API migration territory

Would you be able to tackle this migration @artonge?

Pytal:
If the behaviour in the screencast is not seen, you can instead observe the following error in the console
![image](https://github.com/nextcloud/photos/assets/24800714/3e653e0e-40fb-43ae-9026-0786a4c3d143)

JuliaKirschenheuter:
Thanks @Pytal!

Pytal:
This is a tricky bug as we use focus trap + virtual scrolling here. Focus trap finds the initial tabbable nodes but as you scroll the elements are rerendered and seems to cause the glitch. `updateContainerElements` does not seem to work, still not sure why though.

According to the `focusin` event the next node is correctly focused but immediately afterwards the focus is moved to `body`

Draft semi-working patch to bring back focus to the correct next element in https://github.com/Pytal/focus-trap/pull/1

Pytal:
Regarding the virtual scrolling part, there is a new related bug with the virtual-scrolled files list where if you have many files and tab past the last file to the start of the page and then back into the list it will focus not on the first file but on the first currently rendered file i.e. 90th file row instead of the first

I believe there was some exception made for infinite scrolling before so maybe not an issue

ShGKme:
Can we disable `focus-trap` here and implement our own simple focus trap?

Pytal:
> Can we disable `focus-trap` here and implement our own simple focus trap?

[...]

PR #2256, opened 2024-01-17
fix: Fix focus loop on tab navigation in Add to album modal + Migrate to new UploadPicker API

Pytal:
- For https://github.com/nextcloud/photos/issues/2238

### Summary

- Fix max call stack errors with deps upgrade
- Fix tab navigation loop (not an ideal solution but also a minimally invasive solution for now at least)


Pytal:
Local link works but not package.json 🙇‍♂️

Pytal:
Ok, so the changes between the beta and 1.0 of the upload lib were larger than initially thought 💀

Here a string destination is passed to the UploadPicker component when it is supposed to be a `Folder` https://github.com/nextcloud-libraries/nextcloud-upload/blob/b4b11208fdb60ef133595b5f02a3d9b31384fb99/lib/components/UploadPicker.vue#L132 https://github.com/nextcloud-libraries/nextcloud-files/blob/d3bff085cd6e4a27a87ccafae4c24401715a14da/lib/index.ts#L71 but there is no folder as the "Folder" in this context is an album which does not simply map onto a `Folder` which extends the `Node` type

Current state fixes the "Add photos to this album" action not showing up but direct file upload does not work due to destination not being a Folder

This was initially implemented with upload picker pre-1.0 which allowed non-Folder destinations but 1.0 restricts it to Folders exclusively which is incompatible

What are your thoughts on how we should resolve this @artonge?

artonge:
I guess we need to make a WebDAV request to retrieve the folder's info?
@skjnldsv might have a better idea.

skjnldsv:
> I guess we need to make a WebDAV request to retrieve the folder's info?
> @skjnldsv might have a better idea.

I guess it's because Photos isn't using the Folder/File/Node API?
If so you need to fetch it yes. But you might hack your way into it if you have the weird and old FileInfo. You should be able to generate a somewhat valid Folder objec 🙈

Pytal:
Thanks for the insights @skjnldsv @artonge :)

I have no clue about migrating photos to the Folder/Node API so closing this

artonge:
I have updated `AlbumContent` to pass a `Folder` to the `UploadPicker`.
But I am not sure what to do about the focus trap glitch. Is there a way to refresh the list of focusable elements?

Pytal:
> But I am not sure what to do about the focus trap glitch. Is there a way to refresh the list of focusable elements?

Seems fixed to me @artonge or are you still encounterting the bug?

Pytal:
A bit of migration to the new API left for @artonge

Otherwise ready for review!

artonge (approved):

JuliaKirschenheuter:
Great that it works! Thanks Chris!

Issue #2225, opened 2023-12-31
OriginalDateTimeMetadataProvider unexpected null value


jakkarth:
**Describe the bug**
When generating metadata for a file, an unhandled exception is thrown. OriginalDateTimeMetadataProvider.php on line 90 is assuming that because the array key exists, its corresponding value cannot be null, and passes it as arg 2 to dateToTimestamp.

**To Reproduce**
1. Have a file with some unexpected? metadata in it?
2. Inside the docker container, use occ to generate metadata `php occ files:scan --generate-metadata --all -v --path '/myuser/files/Photos/AFolder/PXL_20231225_235744352.jpg'`

**Expected behavior**
Metadata is properly generated without any errors. `occ` proceeds on to the next file.

**Actual behavior**
Unhandled exception is thrown, metadata isn't generated, stack trace is printed to console, `occ` exits non-zero.

**Stack Trace**
```
An unhandled exception has been thrown:
TypeError: OCA\Photos\Listener\OriginalDateTimeMetadataProvider::dateToTimestamp(): Argument #2 ($date) must be of type string, null given, called in /var/www/html/apps/photos/lib/Listener/OriginalDateTimeMetadataProvider.php on line 90 and defined in /var/www/html/apps/photos/lib/Listener/OriginalDateTimeMetadataProvider.php:49
Stack trace:
#0 /var/www/html/apps/photos/lib/Listener/OriginalDateTimeMetadataProvider.php(90): OCA\Photos\Listener\OriginalDateTimeMetadataProvider->dateToTimestamp('Y:m:d G:i:s', NULL, Object(OC\Files\Node\File))
#1 /var/www/html/lib/private/EventDispatcher/ServiceEventListener.php(86): OCA\Photos\Listener\OriginalDateTimeMetadataProvider->handle(Object(OCP\FilesMetadata\Event\MetadataLiveEvent))
#2 /var/www/html/3rdparty/symfony/event-dispatcher/EventDispatcher.php(230): OC\EventDispatcher\ServiceEventListener->__invoke(Object(OCP\FilesMetadata\Event\MetadataLiveEvent), 'OCP\\FilesMetada...', Object(Symfony\Component\EventDispatcher\EventDispatcher))
[...]

PR #2279, opened 2024-01-24
Just because an array key doesn't imply the value at that key isn't null.

jakkarth:
Fixes nextcloud/photos#2225.


artonge (approved):
Thanks :)

PR #2280, opened 2024-01-24
Filter out large EXIF entries

artonge:


lib/Listener/ExifMetadataProvider.php
```diff
@@ -169,7 +169,10 @@ private function sanitizeEntries(array $data): array {
 				$key = preg_replace('/[^a-zA-Z]/', '_', $key);
 			}
 
-			$cleanData[$key] = $value;
+			// Arbitrary limit to filter out large EXIF entries.
+			if (strlen($value) < 1000) {
+				$cleanData[$key] = $value;
+			}
```
ArtificialOwl, line 175: debug log entries that exceed this limit ? :)

ArtificialOwl (approved):

ArtificialOwl:
why not filters on `UndefinedTag__x____` ?

lib/Listener/ExifMetadataProvider.php
ArtificialOwl: might not need the value itself, just the key and the strlen() of the value ?

artonge:
> why not filters on `UndefinedTag__x____` ?

Today it's `UndefinedTag__x____`, tomorrow, who knows? `UndefinedTag__x____` could also sometime be sane.

Issue #2287, opened 2024-01-26
[BITV]: double scrollbar appears in a file picker dialog


JuliaKirschenheuter:
Double scrollbar appears in a file picker dialog and it is not possible to reach "close" button via mouse

![Screenshot from 2024-01-26 09-57-27](https://github.com/nextcloud/photos/assets/6078378/67f98067-63f8-4d50-b871-c3493632fa32)

**To Reproduce**
1. Go to photos/albums
2. Create new album
3. Click to "Add"
4. Make a viewport a bit smaller (height)

JuliaKirschenheuter:
@susnux @Pytal could it be a fix for this issue too https://github.com/nextcloud/photos/pull/2295?

PR #2296, opened 2024-01-31
enh: Migrate photos picker to use NcDialog

susnux:
* Resolves https://github.com/nextcloud/photos/issues/2287

This simplifies the code and also fixes the double scrollbar issue.


Pytal (commented):
The left nav doesn't show up anymore 🤔

susnux:
> The left nav doesn't show up anymore 🤔

I only enabled the navigation for > 1 entries. Reverted.

Pytal (commented):
image
Padding needs adjustment

src/components/PhotosPicker.vue
```diff
@@ -20,51 +21,26 @@
...
+			<ul>
+				<li v-for="month in monthsList" :key="month">
+					<NcButton :type="targetMonth === month ? 'secondary' : 'tertiary'"
+						:aria-label="t('photos', 'Jump to {date}', { date: dateMonthAndYear(month) })"
+						@click="targetMonth = month">
+						{{ dateMonthAndYear(month) }}
+					</NcButton>
+				</li>
+			</ul>
+		</template>
```
Pytal, line 40: When there are many months and viewport is 320px, it does not scroll

We should use NcSelect when in mobile view like in nc/dialogs FilePicker

Pytal (commented):
When there are many months

src/components/PhotosPicker.vue
susnux: Done :heavy_check_mark: 


localhost_8082_index php_apps_photos_albums_a

susnux:
>  Padding needs adjustment

Done 

Screenshot_20240201_005028

src/components/PhotosPicker.vue
```diff
@@ -20,51 +21,45 @@
...
+		:open="open"
+		out-transition
+		size="large"
+		@update:open="(open) => $emit('update:open', open)">
+		<!-- Navigation containing the months available -->
+		<template #navigation="{ isCollapsed }">
+			<!-- Mobile view -->
+			<NcSelect v-if="isCollapsed"
+				v-model="targetMonth"
+				:aria-label-combobox="t('photos', 'Jump to specific date in list')"
```
Pytal, line 35: We should use the always visible `inputLabel` here

Pytal, line 35: An `aria-label-listbox="t('photos', 'Dates')"` should be added so screen reader users have "Dates list" announced

src/components/PhotosPicker.vue
```diff
@@ -20,51 +21,45 @@
...
-		</div>
-
-		<div class="photos-picker__actions">
+				<template #option="{ label }">
+					{{ dateMonthAndYear(label) }}
+				</template>
+			</NcSelect>
+
+			<!-- Default view -->
+			<ul v-else>
```
Pytal, line 48:
```suggestion
			<ul v-else aria-label="t('photos', 'Dates')">
```
For screen reader users similar to above

JuliaKirschenheuter (approved):

susnux:
@Pytal resolved your comments, please re-review :)

szaimen:
Sorry, I don't feel knowledgeable enough to review this...

artonge (approved):

artonge:
Code looks good, but you probably need to update the cypress tests.

emoral435 (approved):
PR looks fantastic! Reviewed, and tested on my local dev, everything works as intended. 

NcDialog does have this weird look when the close button is focused, though 🤔
image

The active border gets clipped, let me know if you want me to fix this or just think that it is not a high priority issue right now :)

susnux:
> NcDialog does have this weird look when the close button is focused, though 🤔

This should be fixed on the nextcloud-vue-library

Pytal (changes requested):
Tested with a lot of photos and the photo at the end is cut off 🤔
image

Pytal:
Consider this check ✅ my approval in advance once the cut off issue is fixed :D

Pytal:
Cut off issue is still present but just tested master and it also has the same issue so will cc @artonge for investigation at a later point

Pytal (approved):
Works well! Just have to update cypress as mentioned above

PR #2344, opened 2024-02-15
initMetadata on a repair step

ArtificialOwl:
migrate the initMetadata to a repair step process instead of running it too frequently.
the idea is to (server-side) migrate known metadata to lazy appconfig


ArtificialOwl:
_(backport to stable28 is not mandatory)_

artonge (approved):

PR #2428, opened 2024-04-12
fix(3rd-party apps): do not link to maps/recognize install page if the appstore isn't enabled

tcitworld:
Even if the user is admin, they might not have access to the app store.


src/router/index.js
```diff
@@ -51,7 +53,7 @@ const baseTitle = document.title
 Vue.use(Router)
 
 let mapsPath = generateUrl('/apps/maps')
-if (!isMapsInstalled) {
+if (!isMapsInstalled && isAppStoreEnabled) {
```
artonge, line 56: Not sure that it makes sense to have that check here, the link will be hidden anyway, and the link will be wrong anyway too

artonge (approved):

PR #2517, opened 2024-07-04
[stable29] Various fix from 30

artonge:
Manual partial backport of https://github.com/nextcloud/photos/pull/2514


Fenn-CS (commented):
Fewer commits?

Fenn-CS:
Ah I see the partial, what are we leaving out?

artonge:
> Ah I see the partial, what are we leaving out?

Unneeded or unvisible fixes:

- chore: Fix typo in collection.js - not visible
- chore: Use NcSelect label - no error in the browser console
- fix: Don't proceed with album creation if it fails - 29 does not need the extra error handling
- chore: Explicitly require @nc/router - only to make linter happy, so superfluous
- fix: Specify a source for the tmp local album - no error in the browser console

Fenn-CS (approved):

Issue #2259, opened 2024-01-18
[PHP] Error: getimagesizefromstring(): Error reading from ! at /var/www/html/nextcloud/apps/photos/lib/Listener/SizeMetadataProvider.php#56 	from ?


bernd-wechner:
**Describe the bug**
My log file is filled with this message:

[PHP] Error: getimagesizefromstring(): Error reading from ! at /var/www/html/nextcloud/apps/photos/lib/Listener/SizeMetadataProvider.php#56
	from ?

**Screenshots**
Just a small section of the Logging viewer:
![image](https://github.com/nextcloud/photos/assets/7296506/27b586bf-5b4a-41aa-9a4c-fa0ab7fe5231)

**Nextcloud (please complete the following information):**
- Nextcloud-Version: [Nextcloud Hub 7](https://nextcloud.com/) (28.0.1)
- Forms-Version:[4.0.0

**Desktop (please complete the following information):**
 - OS:  Kernel: 5.15.0-91-generic x86_64 bits: 64 Console: pty pts/3,  Distro: Ubuntu 22.04.3 LTS (Jammy Jellyfish)
 - Browser: Firefox
 - Version 121

**Additional context**
A complete formatted message is:

```
[PHP] Error: getimagesizefromstring(): Error reading from ! at /var/www/html/nextcloud/apps/photos/lib/Listener/SizeMetadataProvider.php#56
	from ? by -- at 18 Jan 2024, 10:23:00 pm
```

and a complete raw message is:

```
{
  "reqId": "M10zk8LUgTEAALmH7HB5",
  "level": 3,
  "time": "2024-01-18T22:23:00+11:00",
  "remoteAddr": "",
  "user": "--",
  "app": "PHP",
  "method": "",
  "url": "--",
  "message": "getimagesizefromstring(): Error reading from ! at /var/www/html/nextcloud/apps/photos/lib/Listener/SizeMetadataProvider.php#56",
  "userAgent": "--",
  "version": "28.0.1.1",
  "data": {
    "app": "PHP"
  },
  "id": "65a997949bbeb"
}
```

MihailCosmin:
Have same issue,
[Nextcloud Hub 7] (28.0.1)

![image](https://github.com/nextcloud/photos/assets/14019626/2bb45779-0de9-4f26-89f6-5819d1b5f912)

RobTheBuilder2:
[...]

PR #2524, opened 2024-07-06
fix: avoid erroring out on corrupt image data

### LINKED ISSUE #2235: getimagesizefromstring() : Corrupt JPEG data
**Describe the bug**
Since NC28 upgrade, log file is flooded with this error (about 10 times per minute) :
```getimagesizefromstring(): Corrupt JPEG data: 1 extraneous bytes before marker at /[NC path]/apps/photos/lib/Listener/SizeMetadataProvider.php#56```

I ignored it before seeing long mariaDB queries generating heavy CPU load.

**To Reproduce**
1. Get NC 27 and Photos app fully working
2. Upgrade NC from 27 to 28.0.1
3. See repeated errors in log, and see CPU load and iowait >20%
4. Disable Photos app => back to normal

**Expected behavior**
* No error in log
* Acceptable CPU load and IOwait

**Server log**
```getimagesizefromstring(): Corrupt JPEG data: 1 extraneous bytes before marker at /[NC path]/apps/photos/lib/Listener/SizeMetadataProvider.php#56```

**Additional context**
* Debian Bookworm
* NC 28.0.1
* MariaDB 11.2.2
* PHP 8.2.14

### LINKED ISSUE #2257: getimagesizefromstring(): PNG file corrupted by ASCII conversion at /var/www/html/apps/photos/lib/Listener/SizeMetadataProvider.php#56
**Describe the bug**
Upgrade NC27 to 28, I've got this error flooded in the log.
My NC is running with docker.
```
getimagesizefromstring(): PNG file corrupted by ASCII conversion at /var/www/html/apps/photos/lib/Listener/SizeMetadataProvider.php#56
```

**To Reproduce**
1. Install NC27 and photo app.
2. Upgrade NC28.
3. Run NC at least one day and check the log.

**Expected behavior**

No error in the log

**Screenshots**

msedge_x8za7YgwPj

 - OS: Windows11
 - Browser: MS Edge
 - Version: 120.0.2210.133 (x64)

 - Device: Samsung Galaxy Z Fold 3
 - OS: Android 14
 - Browser: MS Edge
 - Version: 120.0.2210.126

**Additional context**
Running with these docker images
* Nextcloud 28.0.1-apache
* mariadb 10.5
* redis 6-alpine

joshtrichards:
Fixes #2257
Fixes #2235 

As an aside, we should consider switching from using `getimagesizefromstring()` to `getimagesize()` if we can -- it consumes far less memory in my testing (presumably since the entire image isn't loaded into a string in memory). 

P.S. Technically this also addresses the logging side of #2259, but it's unclear whether that's arising from a memory limit or what. I can only get that error if I feed `getimagesizefromstring()` `/dev/null`. :-) However this PR does add logging of the path so it'll likely help work out that issue as well.

### DIFF
--- a/lib/Listener/SizeMetadataProvider.php
+++ b/lib/Listener/SizeMetadataProvider.php
@@ -47,9 +47,10 @@ class SizeMetadataProvider implements IEventListener {
 			return;
 		}
 
-		$size = getimagesizefromstring($node->getContent());
+		$size = @getimagesizefromstring($node->getContent());
 
 		if ($size === false) {
+			$this->logger->debug('SizeMetadataProvider->handle(): Corrupt image data detected ' . $node->getPath());
 			return;
 		}
 


artonge:
> As an aside, we should consider switching from using `getimagesizefromstring()` to `getimagesize()` if we can

I think to recall that `getimagesize()` would not work well with external storage. But maybe I am wrong.

artonge (approved):

PR #2557, opened 2024-07-24
chore: Prepare v3.0.0-beta.1

artonge:


artonge:
No need for that, it's shipped :see_no_evil:

PR #2561, opened 2024-07-27
Migrate to TOML format

AndyScherzinger:
Migrate to `toml` format. 

- [x] Latest scanner now detects the license.txt which need to be handled via @susnux's little helper creating actual .license files
- [x] remove license txt files
- [x] fix unprecise license of photos
- [x] add additional licenses comign in via compiled js dependencies


susnux (approved):

webpack.js
```diff
@@ -87,4 +88,17 @@ webpackConfig.plugins.push(
...
+		override: {
+			// TODO: Remove if they fixed the license in the package.json
+			'@nextcloud/axios': 'GPL-3.0-or-later',
+			'@nextcloud/vue': 'AGPL-3.0-or-later',
+			'nextcloud-vue-collections': 'AGPL-3.0-or-later',
+		}
+	}),
+]
+
 module.exports = webpackConfig
```
susnux, line 104: still needs to disable the extract comments part of the terser minifier to remove the `.license.txt` files

AndyScherzinger: 👍
Yes, just need to find the config snippet again 😅

AndyScherzinger: found it 😃

Issue #2635, opened 2024-09-06
ExifMetadataProvider.php Doesn't Handle Unexpected Data Well


nmbgeek:
**Describe the bug**
Parsing of GPS data needs to be improved for files with various metadata issues

**To Reproduce**
Upload a file with missing or non-standard GPS metadata

**Expected behavior**
Handle the missing metadata gracefully

**Desktop (please complete the following information):**
- Server side error and UI hangs.

<details>
#2541 contains an image that will hang the UI when uploading due to non-standard metadata
#2260 - The main issue may or may not be addressed by the pull request I will be uploading but many of the other respondents will be.

My error which is mentioned under #2260 but not identical to the main issue there.
```
[...]

Issue #2260, opened 2024-01-18
GenericFileException Error while generating metadata


bernd-wechner:
**Describe the bug**
Error messages in the log file. This message is repeated time and again and seems to relate to one file by id (not sure how to identify which file by id at present). 

```
[no app in context] Warning: Error while generating metadata for fileid 42374
	from ? by -- at 18 Jan 2024, 10:09:33 pm
```

and the raw message:

```
{
  "reqId": "M10zk8LUgTEAALmH7HB5",
  "level": 2,
  "time": "2024-01-18T22:09:33+11:00",
  "remoteAddr": "",
  "user": "--",
  "app": "no app in context",
  "method": "",
  "url": "--",
  "message": "Error while generating metadata for fileid 42374",
  "userAgent": "--",
  "version": "28.0.1.1",
  "exception": {
    "Exception": "OCP\\Files\\GenericFileException",
    "Message": "",
    "Code": 0,
    "Trace": [{
      "file": "/var/www/html/nextcloud/apps/photos/lib/Listener/SizeMetadataProvider.php",
      "line": 56,
      "function": "getContent",
      "class": "OC\\Files\\Node\\File",
      "type": "->"
    }, {
      "file": "/var/www/html/nextcloud/lib/private/EventDispatcher/ServiceEventListener.php",
      "line": 86,
      "function": "handle",
      "class": "OCA\\Photos\\Listener\\SizeMetadataProvider",
      "type": "->"
    }, {
      "file": "/var/www/html/nextcloud/3rdparty/symfony/event-dispatcher/EventDispatcher.php",
      "line": 230,
      "function": "__invoke",
      "class": "OC\\EventDispatcher\\ServiceEventListener",
      "type": "->"
    }, {
      "file": "/var/www/html/nextcloud/3rdparty/symfony/event-dispatcher/EventDispatcher.php",
      "line": 59,
      "function": "callListeners",
      "class": "Symfony\\Component\\EventDispatcher\\EventDispatcher",
      "type": "->"
    }, {
      "file": "/var/www/html/nextcloud/lib/private/EventDispatcher/EventDispatcher.php",
      "line": 94,
      "function": "dispatch",
      "class": "Symfony\\Component\\EventDispatcher\\EventDispatcher",
      "type": "->"
    }, {
      "file": "/var/www/html/nextcloud/lib/private/EventDispatcher/EventDispatcher.php",
      "line": 106,
      "function": "dispatch",
[...]

PR #2636, opened 2024-09-06
Fix issues of GPS metadata handling

nmbgeek:
#2541 #2635 #2616 Directly address these issues

#2260 Address some of the issues in the comments

This pull request adds handling null GPS coordinates and also coordinates that are not a string.

### DIFF
--- a/lib/Listener/ExifMetadataProvider.php
+++ b/lib/Listener/ExifMetadataProvider.php
@@ -95,8 +95,8 @@ class ExifMetadataProvider implements IEventListener {
 				$gps['longitude'] = $this->gpsDegreesToDecimal($rawExifData['GPS']['GPSLongitude'], $rawExifData['GPS']['GPSLongitudeRef']);
 			}
 
-			if (array_key_exists('GPSAltitude', $rawExifData['GPS']) && array_key_exists('GPSAltitudeRef', $rawExifData['GPS'])) {
-				$gps['altitude'] = ($rawExifData['GPS']['GPSAltitudeRef'] === "\u{0000}" ? 1 : -1) * $this->parseGPSData($rawExifData['GPS']['GPSAltitude']);
+			if (array_key_exists('GPSAltitude', $rawExifData['GPS']) && array_key_exists('GPSAltitudeRef', $rawExifData['GPS']) && is_string($rawExifData['GPS']['GPSAltitude'])) {
+				$gps['altitude'] = ($rawExifData['GPS']['GPSAltitudeRef'] === '1' || $rawExifData['GPS']['GPSAltitudeRef'] === "\u{0001}" ? -1 : 1) * $this->parseGPSData($rawExifData['GPS']['GPSAltitude']);
 			}
 
 			if (!empty($gps)) {


nmbgeek:
I believe this should also fix #2616 because $rawData can be an array.

lib/Listener/ExifMetadataProvider.php
```diff
@@ -109,11 +109,15 @@ public function handle(Event $event): void {
 	 * @param array|string $coordinates
 	 */
 	private function gpsDegreesToDecimal($coordinates, ?string $hemisphere): float {
+		if (is_null($coordinates)) {
+			throw new \Exception('Coordinates are null');
+		}
+
 		if (is_string($coordinates)) {
-			$coordinates = array_map('trim', explode(',', $coordinates));
+			$coordinates = array_map("trim", explode(",", $coordinates));
```
artonge, line 117: Our coding standard asks for the original :)

lib/Listener/ExifMetadataProvider.php
```diff
@@ -123,7 +127,15 @@ private function gpsDegreesToDecimal($coordinates, ?string $hemisphere): float {
 		return $sign * ($degrees + $minutes / 60 + $seconds / 3600);
 	}
 
-	private function parseGPSData(string $rawData): float {
+	private function parseGPSData($rawData): float {
```
artonge, line 130:
```suggestion
	/**
	 * @param array|string $rawData
	 */
	private function parseGPSData($rawData): float {
```

lib/Listener/ExifMetadataProvider.php
```diff
@@ -123,7 +127,15 @@ private function gpsDegreesToDecimal($coordinates, ?string $hemisphere): float {
...
+		if (is_array($rawData)) {
+		if (count($rawData) === 3) {
+			return floatval($rawData[0]) + floatval($rawData[1]) / 60 + floatval($rawData[2]) / 3600;
+			} else {
+				return 0;  // Handle unexpected array structure
+			}
+		}
+
```
artonge, line 138: Weird indentation.

artonge (commented):
Looks good overall. Minor comment on code styling.
Thanks for looking into it :heart:

lib/Listener/ExifMetadataProvider.php
nmbgeek: I think these are all fixed now.

lib/Listener/ExifMetadataProvider.php
```diff
@@ -122,8 +126,19 @@ private function gpsDegreesToDecimal($coordinates, ?string $hemisphere): float {
...
-
-	private function parseGPSData(string $rawData): float {
+	
+	/**
+	 * @param array|string $rawData
+	 */
+	private function parseGPSData($rawData): float {
+		if (is_array($rawData)) {
+			if (count($rawData) === 3) {
+				return floatval($rawData[0]) + floatval($rawData[1]) / 60 + floatval($rawData[2]) / 3600;
```
artonge, line 136: Why are we doing that computation here? It looks like the coordinate computation that we are doing above.
Could you add some comment explaining the different formats?

nmbgeek: After reviewing the correct implementation of EXIF metadata for altitude you are correct, there is no need for this.  Apparently some cameras/applications add invalid metadata.  Altitude in only parsed with this parseGPSData and I was trying to correct for that.  I believe just adding a string check to line 98 may be the better solution?

`if (array_key_exists('GPSAltitude', $rawExifData['GPS']) && array_key_exists('GPSAltitudeRef', $rawExifData['GPS']) && is_string($rawExifData['GPS']['GPSAltitude'])) {`

artonge: If this only occur with `GPSAltitude`, then yes. But I am wondering if we could handle the array format. Maybe add the `if` condition to prevent the error, but also log the value so we can have more information to improve the handling in the future?

nmbgeek: #2541 Has an image that has Altitude as '0 0 0' and breaks without checking if it is a string.  The GPS Data also appears to have been scrubbed from the image instead of just removing it with latitude and longitude also both being 0.  From reading some exif documentation it should just be a rational64u formatted as numerator/denominator.  If the data isn't valid is there really anything else to do with it?  It doesn't appear that Photos is actually doing anything with the altitude value either.

lib/Listener/ExifMetadataProvider.php
```diff
@@ -95,9 +95,9 @@ public function handle(Event $event): void {
 				$gps['longitude'] = $this->gpsDegreesToDecimal($rawExifData['GPS']['GPSLongitude'], $rawExifData['GPS']['GPSLongitudeRef']);
 			}
 
-			if (array_key_exists('GPSAltitude', $rawExifData['GPS']) && array_key_exists('GPSAltitudeRef', $rawExifData['GPS'])) {
-				$gps['altitude'] = ($rawExifData['GPS']['GPSAltitudeRef'] === "\u{0000}" ? 1 : -1) * $this->parseGPSData($rawExifData['GPS']['GPSAltitude']);
-			}
+			if (array_key_exists('GPSAltitude', $rawExifData['GPS']) && array_key_exists('GPSAltitudeRef', $rawExifData['GPS']) && is_string($rawExifData['GPS']['GPSAltitude'])) {
+				$gps['altitude'] = ($rawExifData['GPS']['GPSAltitudeRef'] === '1' ? -1 : 1) * $this->parseGPSData($rawExifData['GPS']['GPSAltitude']);
```
artonge, line 99: Is this change necessary ?

lib/Listener/ExifMetadataProvider.php
```diff
@@ -95,9 +95,9 @@ public function handle(Event $event): void {
...
+			}
```
artonge, line 100: Extra white space

lib/Listener/ExifMetadataProvider.php
nmbgeek: This was originally `$rawExifData['GPS']['GPSAltitudeRef'] === "\u{0000}" ? 1 : -1)` \u{0000} is supposed to be the null character but the specifications say it can also be 0 for above sea level. A value of 1 indicates below sea level.

artonge: Hmm, alright then. Have you tested it?

nmbgeek: Yes. I tested with this code and expected values were logged.


```
			if (array_key_exists('GPSAltitude', $rawExifData['GPS']) && array_key_exists('GPSAltitudeRef', $rawExifData['GPS']) && is_string($rawExifData['GPS']['GPSAltitude'])) {
				$this->logger->info('GPSAltitude: ' . $rawExifData['GPS']['GPSAltitude']);
				$this->logger->info('GPSAltitudeRef: ' . $rawExifData['GPS']['GPSAltitudeRef']);
				$gps['altitude'] = ($rawExifData['GPS']['GPSAltitudeRef'] === '1' ? -1 : 1) * $this->parseGPSData($rawExifData['GPS']['GPSAltitude']);
				$this->logger->info('GPSAltitudeCalc: ' . $gps['altitude']);
			}
```

nmbgeek: I tested with a photo that I used exiftool to write metadata for below sea level and it didn't work.  The logged value for GPSAltitudeRef was "\u0001" so I will revert it to using the unicode character but assuming above sea level if the value isn't explicitly below.

artonge (approved):

artonge:
@nmbgeek can you squash your changes?

artonge:
My bad, I rebased your branch to run the CI against the latest master. This should help you:

```shell
git reset --hard HEAD~11
git cherry-pick b674472a58b9fb2535374f3cb4e72e6155e9c563
git push --force-with-lease
```

PR #2657, opened 2024-09-25
feat: Check upload directory share state

artonge:
Done:
- Block upload to the album DAV endpoint if the current user does not own the photo folder.
- Add warning in Photos picker dialog. An alternative would be to hide the button. But uploads cannot be blocked as they are done directly to the DAV files endpoint, so it is impossible to distinguish from a legitimate upload. 

Screenshot from 2024-09-25 14-40-42

### DIFF
--- a/lib/Sabre/Album/AlbumRoot.php
+++ b/lib/Sabre/Album/AlbumRoot.php
@@ -14,6 +14,7 @@ use OCA\Photos\Album\AlbumMapper;
 use OCA\Photos\Album\AlbumWithFiles;
 use OCA\Photos\Service\UserConfigService;
 use OCP\Files\Folder;
+use OCP\Files\InvalidDirectoryException;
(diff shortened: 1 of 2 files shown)


lib/Sabre/Album/AlbumRoot.php
```diff
@@ -81,6 +82,11 @@ public function createFile($name, $data = null) {
 				throw new Conflict('The destination exists and is not a folder');
 			}
 
+			if ($photosFolder->isShared()) {
+				throw new Conflict('The destination exists and is not a folder');
+				throw new InvalidDirectoryException('The destination is a received share');
```
nickvergessen, line 87: Can only throw once

provokateurin (approved):

artonge:
Closed in favor of local copy: https://github.com/nextcloud/photos/pull/2658

PR #2740, opened 2024-12-30
chore: update composer lock and nextcloud/ocp

skjnldsv:
Fix https://github.com/nextcloud/photos/issues/2738


artonge (approved):

skjnldsv:
Don't think the cypress failure is related

PR #2794, opened 2025-02-03
chore: Move Create new album action out of menu

Pytal:
image


Pytal:
CI failure unrelated but keeps failing :(

Can you take a look @artonge? 🙏

PR #2813, opened 2025-02-19
chore: Bump php version in psalm.xml and composer.json

artonge:


artonge:
Included in https://github.com/nextcloud/photos/pull/2766#

PR #2928, opened 2025-04-10
chore: Migrate to vite

artonge:
Minimal changes to switch to vite.
Better reviewed per commit.


package.json
```diff
@@ -90,29 +91,24 @@
     "@nextcloud/cypress": "^1.0.0-beta.12",
```
susnux, line 91: babel config above not needed anymore

skjnldsv (approved):

package.json
susnux: (if cypress is also using vite)

susnux (approved):

PR #2949, opened 2025-04-23
perf: group icons into one chunk and extract CSS

susnux:
- group all icons into one chunk -> reduce from 73 to 56 files
- extract CSS to reduce JS parser time


REUSE.toml
```diff
@@ -12,7 +12,13 @@ SPDX-FileCopyrightText = "none"
 SPDX-License-Identifier = "CC0-1.0"
 
 [[annotations]]
-path = ["l10n/**.js", "l10n/**.json", "js/**.js.map", "js/**.js", "js/**.mjs", "js/**.mjs.map", "js/templates/**.handlebars", "screenshots/screenshot1.png", "src/assets/file-placeholder.svg"]
+path = ["js/**.map"]
+precedence = "override"
```
susnux, line 16: This is needed to remove warning from REUSE because there are SPDX headers within the `map` file (as they contain all sources).

It basically tells REUSE to use this license here instead of parsing the file.

artonge (approved):

skjnldsv (commented):
Still not sure about the benefits of 36 css files :see_no_evil:

PR #2963, opened 2025-04-23
chore: fix some Typescript issues

susnux:
Fix some Typescript issues I noticed while testing the other pull requests.


artonge:
Might be merged after https://github.com/nextcloud/photos/pull/2945. But don't want to have to deal with conflict for now :).

PR #2967, opened 2025-04-24
perf: Further group output files

artonge:
Screenshot 2025-04-28 at 11-24-45 Vite Bundle Visualizer


vite.config.ts
```diff
@@ -28,6 +28,28 @@ export default createAppConfig({
...
+						}
+						if (id.toLowerCase().includes('tag')) {
+							return 'tags'
+						}
+						if (id.toLowerCase().includes('face')) {
+							return 'faces'
+						}
+						if (id.toLowerCase().includes('sidebar') || id.toLowerCase().includes('tab') || id.toLowerCase().includes('leaflet')) {
+							return 'photos-sidebar'
+						}
```
susnux, line 48: This is not quite good I would argue because you mix here also internal modules, e.g. if there is a node module with "tag" inside the name you will group it in "tags" which will make the chunking useless.

vite.config.ts
```diff
@@ -28,6 +28,28 @@ export default createAppConfig({
...
+						if (id.toLowerCase().includes('node_modules')) {
+							return 'vendor'
+						}
```
susnux, line 51: Especially this I would consider not that good as then you do not even need to create any different entry points at all. Because then you again bundle everything in one chunk even if it is never used by the entry point.

vite.config.ts
```diff
@@ -28,6 +28,28 @@ export default createAppConfig({
...
+						return 'shared'
```
susnux, line 52: Same here

susnux (commented):
Not sure this is really something "performance".

You sacrifice parsing performance which is done **everytime** someone opens the photos app for the **one time** performance of a couple of requests more which are cached.
Also those requests are done async partially after first load so even not as bad for high latency.

I personally would not mess that much with internal chunking logic, as then you could also just remove any chunking at all.

susnux (commented):
I just spotted your screenshot, thats horrible :fearful: 

Because:
- No manual chunking:
  - Many smaller chunks as everything is separated so it is only imported once
  - Meaning small chunks which are fast to parse
  - No useless code is interpreted
  - But more HTTP request on **first** page load
- Making only few chunks:
  - Few HTTP requests which lower waiting time on **bad** networks
  - Includes a lot of useless code as its not separated by whats needed for entry point anymore


So the benefit of specify manual chunks is to support users with bad internet connection where the request has a high latency.
But bad internet connection also means high package drop, so a chunk that large will likely fail a lot and needs much longer to be fully transferred.

Also a chunk that big will make every page load much slower as the full script needs to be parsed, which you can even measure on somewhat recent hardware.

On the other hand small chunks are not that bad as they are cached anyways by the browser and in theory the web server can already merge a lot of files into one request when using HTTP2.

artonge:
It was an attempt to group small chunks that would eventually be needed together, but I agree that the result is not improving the situation.

PR #3001, opened 2025-05-06
fix: Adapt sidebar tab to new fileinfo structure

artonge:


src/views/PhotosTab.vue
```diff
@@ -92,7 +92,7 @@ export default defineComponent({
 			}
 		},
 		originalDateTime(): number {
-			return this.fileInfo['metadata-photos-original_date_time'] * 1000
+			return this.fileInfo?.attributes['metadata-photos-original_date_time'] * 1000
```
susnux, line 95: Can it be undefined? If yes than this is invalid code.

If `this.fileInfo` can be undefined then this would be `undefined * 1000` which is **not** `0` but `NaN`.
If it cannot be undefined then use `!`.

So either:
```suggestion
			return (this.fileInfo?.attributes['metadata-photos-original_date_time'] ?? 0) * 1000
```

or
```suggestion
			return this.fileInfo!.attributes['metadata-photos-original_date_time'] * 1000
```

susnux: (same on the other places)

PR #3003, opened 2025-05-07
feat: Support setting and getting filters in albums

artonge:


artonge:
Can you run `composer i` inside photos?

Antreesy:
That helped, thanks! That's a local-only history, package users / customers shouldn't later stuck on it?

PR #3015, opened 2025-05-11
chore(i18n): Improved grammar

rakekniven:


src/views/Timeline.vue
```diff
@@ -134,7 +134,7 @@
 
 		<NcModal v-if="showAlbumPicker"
 			key="albumPicker"
-			:label-id="t('photos', 'album-picker')"
+			:label-id="t('photos', 'album picker')"
```
artonge, line 137: Should this even be translated? I think not.
```suggestion
			label-id="album-picker"
```

artonge (changes requested):

src/views/Timeline.vue
artonge: It is a technical property of the component, so it is not rendered.

artonge (approved):
❤️

PR #3018, opened 2025-05-15
chore: Apply rector php8.1 set

artonge:

### DIFF
--- a/lib/Command/AlbumAddCommand.php
+++ b/lib/Command/AlbumAddCommand.php
@@ -17,16 +17,12 @@ use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Output\OutputInterface;
 
 class AlbumAddCommand extends Command {
-	private readonly IRootFolder $rootFolder;
-	private readonly IUserManager $userManager;
 
 	public function __construct(
 		private readonly AlbumMapper $albumMapper,
-		IRootFolder $rootFolder,
-		IUserManager $userManager,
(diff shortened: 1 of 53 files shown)


lib/Command/AlbumAddCommand.php
```diff
@@ -17,18 +17,16 @@
...
-	private IUserManager $userManager;
-	private AlbumMapper $albumMapper;
+	private readonly IRootFolder $rootFolder;
+	private readonly IUserManager $userManager;
 
 	public function __construct(
-		AlbumMapper $albumMapper,
+		private readonly AlbumMapper $albumMapper,
 		IRootFolder $rootFolder,
 		IUserManager $userManager,
```
come-nc, line 26: Somehow it missed these ones :thinking:

lib/Command/AlbumCreateCommand.php
```diff
@@ -17,15 +17,13 @@
...
 
 class AlbumCreateCommand extends Command {
-	private IUserManager $userManager;
-	private AlbumMapper $albumMapper;
+	private readonly IUserManager $userManager;
 
 	public function __construct(
-		AlbumMapper $albumMapper,
+		private readonly AlbumMapper $albumMapper,
 		IUserManager $userManager,
```
come-nc, line 24: Same here

come-nc (commented):
So weird that it’s refusing to use promoted constructor properties for some classe :thinking:

PR #3030, opened 2025-05-21
[stable31] fix(perf): Do not request faces properties in search requests

artonge:


susnux (approved):

susnux:
Makes sense, in general the current way with nc files is too broad, we need also for the files app not only one way of requesting props but to have standard props and only-per-request props.
E.g. the rich workspace is such a property.

PR #3041, opened 2025-05-27
fix(photos): validate source folders in PhotoSearch

nfebe:
- Add validation to ensure photosSourceFolders is an array
- Filter out invalid folder paths (non-string or empty)
- Add error handling for path joining
- Return early if no valid source folders are found
- Improve error logging for debugging

This fixes an issue where invalid source folder configurations would cause the app to crash with a "lastArg.charAt is not a function" error.


artonge:
Uneeded

PR #3042, opened 2025-05-27
fix: Propagate DAV permissions to the Viewer

artonge:
Follow-up of the migration to the Node API, this was missing to allow users to edit pictures from the Photos app.


src/utils/fileUtils.ts
```diff
@@ -79,6 +79,14 @@ export function sortCompareFileInfo(fileInfo1: FoldersNode, fileInfo2: FoldersNo
...
 
 export function toViewerFileInfo(file: Node) {
+	let permissions = ''
+
+	if ((file.permissions & Permission.CREATE) === Permission.CREATE) { permissions += 'CK' }
+	if ((file.permissions & Permission.UPDATE) === Permission.UPDATE) { permissions += 'WNV' }
+	if ((file.permissions & Permission.READ) === Permission.READ) { permissions += 'G' }
+	if ((file.permissions & Permission.DELETE) === Permission.DELETE) { permissions += 'D' }
+	if ((file.permissions & Permission.SHARE) === Permission.SHARE) { permissions += 'R' }
+
```
nfebe, line 89: What do these shorthand codes mean?

nfebe (approved):
Works

PR #3147, opened 2025-09-02
Replace mdi download icon with Material Symbol variant

AndyScherzinger:


AndyScherzinger:
should be fixed now @susnux but have to do it manually, can run it locally on my windows setup

PR #3202, opened 2025-09-27
Add exif support for webp images

balping:
PHP's built-in `exif_read_data()` does not support webp files natively at the moment. As a workaround, [webp-exif](https://github.com/WoltLab/webp-exif) can be used to extract exif tags.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -38,7 +38,8 @@
 		"nextcloud/ocp": "dev-master"
 	},
 	"require": {
-		"hexogen/kdtree": "^0.2.5"
+		"hexogen/kdtree": "^0.2.5",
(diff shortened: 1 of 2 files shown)


artonge (approved):

artonge:
> @artonge I pushed composer.lock now

Thanks, this is needed so that we all share the same dependencies versions.

Can you handle the conflicts?

artonge:
And please rebase on top of master instead of pulling changed from it to avoid creating a merged commit.

```bash
git reset --hard HEAD~1
git rebase origin/master
```

balping:
@artonge I don't know why it's failing now. Running psalm locally doesn't give errors.

Issue #3205, opened 2025-09-29
People tab broken in Nextcloud 32


SeekDaSky:
**Describe the bug**
Since the upgrade to nextcloud 32 I cannot navigate to the "People" tab. Looking at the console I see

```
ReferenceError: RouterLink is not defined
    <anonymous> FacesView.vue:63
```

I tried to "clean install" everything (re-installed all the nextcloud files, re-install the `recognize` app) but the problem persists.

**To Reproduce**
1. Have the "recognize" app installed
2. Try to navigate to the "People" tab
3. Nothing happens and there is an error in the console

**Expected behavior**
I should be able to go to the people tab

**Desktop (please complete the following information):**
 - OS: MacOS 15.7
 - Browser Firefox
 - Version 143

**Browser log**
```
ReferenceError: RouterLink is not defined
    <anonymous> FacesView.vue:63
```

<details>
<summary>How to access your browser console (Click to expand)</summary>

# Chrome
- Press either CTRL + SHIFT + J to open the “console” tab of the Developer Tools.
- Alternative method:
    1. Press either CTRL + SHIFT + I or F12 to open the Developer Tools.
    2. Click the “console” tab.

# Safari
- Press CMD + ALT + I to open the Web Inspector.
- See Chrome’s step 2. (Chrome and Safari have pretty much identical dev tools.)

# IE9
1. Press F12 to open the developer tools.
2. Click the “console” tab.

# Firefox
- Press CTRL + SHIFT + K to open the Web console (COMMAND + SHIFT + K on Macs).
- or, if Firebug is installed (recommended):
    1. Press F12 to open Firebug.
    2. Click on the “console” tab.

# Opera
1. Press CTRL + SHIFT + I to open Dragonfly.
2. Click on the “console” tab.
</details>

**Additional context**

PR #3237, opened 2025-10-15
fix(FacesView): Properly import RouterLink

artonge:
Cleaned version of https://github.com/nextcloud/photos/pull/3216
Fix https://github.com/nextcloud/photos/issues/3205


susnux (approved):
Alternative: Do not import it at all. Its globally available when vue router is registered to Vue :)

PR #3270, opened 2025-11-07
fix(albums): Do not overwrite albums on move

artonge:
Else, the user completely lose the target album and its content.


artonge:
@tobiasKaminsky ir was done from 31: https://github.com/nextcloud/photos/pull/3273

PR #3317, opened 2025-11-19
chore(queries): avoid left join when possible

Altahrim:


lib/Album/AlbumMapper.php
```diff
@@ -449,10 +449,9 @@ public function getSharedAlbumsForCollaborator(string $collaboratorId, int $coll
 		$rows = $query
 			->select('a.album_id', 'name', 'user', 'location', 'created', 'last_added_photo', 'filters')
 			->from('photos_albums_collabs', 'c')
-			->leftJoin('c', 'photos_albums', 'a', $query->expr()->eq('a.album_id', 'c.album_id'))
+			->innerJoin('c', 'photos_albums', 'a', $query->expr()->eq('a.album_id', 'c.album_id'))
 			->where($query->expr()->eq('collaborator_id', $query->createNamedParameter($collaboratorId)))
 			->andWhere($query->expr()->eq('collaborator_type', $query->createNamedParameter($collaboratorType, IQueryBuilder::PARAM_INT)))
-			->andWhere($query->expr()->isNotNull('a.album_id'))
```
artonge: Probably a left over from copying and pasting as `album_id` should always be set in the `albums` table.
So should be a safe cleanup :+1:

PR #3325, opened 2025-11-19
feat(Albums): Prompt before deleting

artonge:


src/store/collections.ts
```diff
@@ -332,16 +332,63 @@ const actions = {
...
+					confirmTitle = t('photos', 'Delete album')
+					break
+				case 'sharedalbums':
+					confirmTitle = t('photos', 'Leave shared album')
+					break
+			}
+			const result = await confirmOperation(
+				confirmTitle,
+				t('photos', 'Are you sure you want to delete {collectionName}? This action cannot be undone.', { collectionName }),
+			)
```
susnux: If you already use dialogs v7 then you can simplify with `showConfirmation`

susnux (approved):

PR #3327, opened 2025-11-20
feat(Settings): Refine settings components

artonge:
1. Use full width button for sources
2. Merge upload folders info into a single button
3. Re-arrange setting groups and order.

| Before | After |
|--------|--------|
| image | image |


kra-mo (approved):
Nice, thank you :)

Some smaller comments/suggestions:
- It would be nice if the icon for choosing a folder would be consistent with the one in Talk etc.
- The text beneath Media folders is now misaligned. Is this due to a library update? Regardless, it should be fixed.
- In any case, I feel like all settings could be reorganized under just one "General" pane since the app doesn't have a lot. This is for the future, where if an app only has one main settings section, the navigation sidebar will be hidden.
- I would also make some wording changes.

Image

kra-mo:
Oh actually, `showNavigation` is an existing property. It can just manually be set to false if there is only one section, I suppose.

artonge:
> Oh actually, `showNavigation` is an existing property. It can just manually be set to false if there is only one section, I suppose.

So the idea was to hide the sidebar, not to remove the sections titles?

kra-mo (changes requested):
Thank you, it looks really nice now :)

I only have this comment about the period, and there is the hiding of the sidebar.

PR #3436, opened 2026-02-26
fix(faces): Retrieve recognize API key and send it along with DAV requests

### LINKED ISSUE #3200: Heads up: Recognize WebDAV API is changing
Hello 👋 

The photos app uses the recognize WebDAV API for displaying the recognized face clusters in the UI. Due to the EU AI Act, we have to limit access to this API and have opted for the following change, starting with Nextcloud 33:

* Every Nextcloud App that wants to use the Recognize WebDAV API needs to generate an API key in the PHP backend by using the following method: `OCA\Recognize\Public\ApiKeyManager#generateApiKey():string throws \JsonException`
* When sending WebDAV requests to the Recognize WebDAV API, this key must be set as `X-Recognize-Api-Key` header. Each generated key is valid for one day. I recommend using initial state to send the API key to the client.

Thank you for your cooperation.

marcelklehr:
fixes #3200


julien-nc (commented):
Assuming the dependency injector will provide `null` if `\OCA\Recognize\Public\ApiKeyManager` does not exist, these changes look good.

marcelklehr:
mmh, cypress failures seem unrelated:

```
 
  Sidebar tab
    1) Should display EXIF data in the sidebar
    2) Should update EXIF data when changing photos


  0 passing (22s)
  2 failing

  1) Sidebar tab
       Should display EXIF data in the sidebar:
     AssertionError: Timed out retrying after 4000ms: Expected to find element: `#tab-button-photos`, but never found it.
      at Context.eval (webpack://photos/./cypress/e2e/sidebar.cy.ts:25:0)

  2) Sidebar tab
       Should update EXIF data when changing photos:
     AssertionError: Timed out retrying after 4000ms: Expected to find element: `#tab-button-photos`, but never found it.
      at Context.eval (webpack://photos/./cypress/e2e/sidebar.cy.ts:36:0)

```

PR #3476, opened 2026-04-21
Fix downloading files from public share

### LINKED ISSUE #3442: [Bug]: Video playback fails (503) on publicly shared albums - null user ID in DAV URL
## Bug description

When a Photos album is shared via public link and a visitor tries to play a video, the request fails with HTTP 503. Images in the same album work fine.

## Steps to reproduce

1. Create an album in the Photos app containing videos (mp4)
2. Share the album via public link
3. Open the public share URL in a browser (without being logged in)
4. Try to play a video

## Expected behavior

Video plays normally, same as when logged in.

## Actual behavior

Video fails to load with HTTP 503. The browser makes requests to two URLs:

```
GET /public.php/dav/photospublic/{token}/video.mp4  → 503
GET /public.php/dav/files/null/{token}/video.mp4    → 503
```

Note the `null` user ID in the second URL — the Photos app cannot resolve the album owner when serving the request via a public share token, resulting in a broken DAV path.

The Nextcloud application log shows:

```json
{
  "level": 3,
  "app": "no app in context",
  "method": "GET",
  "url": "/public.php/dav/photospublic/{token}/video.mp4",
  "message": "Exception thrown: Sabre\\DAV\\Exception\\NotFound",
  "exception": {
    "Exception": "Sabre\\DAV\\Exception\\NotFound",
    "Message": "",
    "File": "/var/www/html/apps/dav/lib/Connector/Sabre/PublicAuth.php",
    "Line": 133
  }
}
```

`PublicAuth::checkToken()` throws `NotFound` because album share tokens are not stored in the same share backend as regular file shares.

As a side effect, this triggers the brute-force protection, causing subsequent requests to be throttled (HTTP 429) even after the rate limit is reset.

## Workaround

None found for album shares. Sharing the folder directly via the Files app (standard link share) works correctly for video streaming.

## Environment

- Nextcloud version: 33.0.0
- Photos app version: 6.0.0-dev.0 (shipped with `nextcloud:33.0.0-apache` Docker image)
- Reverse proxy: Traefik v3
- Browser: Chrome 145 (Linux)

danxuliu:
Fixes #3442, which is a regression introduced in https://github.com/nextcloud/photos/commit/4a65c8c1b7b8f5016d1189a0f4eb403be77bfaee

When a file is open in the viewer [the Photos app provides an object with the file information](https://github.com/nextcloud/photos/blob/02ebcc1ec51812e715e7b6390cd41bb9a4b65e89/src/components/Collection/CollectionContent.vue#L137). The object includes the `source` attribute, which [is used to get the URL to download the file open in the viewer](https://github.com/nextcloud/viewer/blob/8fbc16a85ac282ee10903d1f0e076e0698204752/src/views/Viewer.vue#L1243).

Before 4a65c8c1b7b8f5016d1189a0f4eb403be77bfaee the `source` was created by appending the file name (`/photospublic/TOKEN/FILE`) to the remote DAV URL (`https://DOMAIN/remote.php/dav`). [In 4a65c8c1b7b8f5016d1189a0f4eb403be77bfaee](https://github.com/nextcloud/photos/commit/4a65c8c1b7b8f5016d1189a0f4eb403be77bfaee#diff-6d05e18efa847b239fa80996f1eed75e9618750902846b35988932aedd28440cR91-R94) instead of the remote DAV URL defined in the app the default remote URL provided by `@nextcloud/files/dav` is used, which [in the case of public pages is `https://DOMAIN/public.php/dav`](https://github.com/nextcloud-libraries/nextcloud-files/blob/508dee00534d5dde63b4b5bc38cf773bbbd66409/lib/dav/dav.ts#L45-L61). However, the file name is not just appended to the default remote URL, but generated through a URL object. As [the file name is relative to the root only the site part of the base URL is used](https://developer.mozilla.org/en-US/docs/Web/API/URL_API/Resolving_relative_references#root_relative) (in this case, the default remote URL), so the final "assembled" URL becomes `https://DOMAIN/photospublic/TOKEN/file`.

A bit later, in 999d49b3acd88a45fc6a3fbdaabf5345dd39cd95, [`genFileInfo` was replaced with `resultToNode`](https://github.com/nextcloud/photos/commit/999d49b3acd88a45fc6a3fbdaabf5345dd39cd95#diff-37aba23b3feb7ee8053792a0242707a92be255702b1146bf84d0c63d5d93c1bfL151-R152) from `@nextcloud/files/dav`, which composes the `source` attribute by appending the file name to the default remote URL, so it becomes `https://DOMAIN/public.php/dav/photospublic/TOKEN/FILE`. While it is now using a "valid" Nextcloud URL the shared files can not be downloaded from it.

The problem with `/public.php/dav` is that it can be used only for "normal" public shares, as it uses a custom DAV server that [does not load the additional authentication backends](https://github.com/nextcloud/server/blob/5dbbe2ccf9570e226f47f89a86204b5a9546264b/apps/dav/appinfo/v2/publicremote.php) and [always uses the public DAV authentication](https://github.com/nextcloud/server/blob/5dbbe2ccf9570e226f47f89a86204b5a9546264b/apps/dav/appinfo/v2/publicremote.php#L59-L68), which [treats the token as a share token and therefore fails](https://github.com/nextcloud/server/blob/9c98b722f4fc4076a56c3cdc1a70b001889ce396/apps/dav/lib/Connector/Sabre/PublicAuth.php#L130). On the other hand, `/remote.php/dav` uses a full DAV server, which [loads the additional authentication backends](https://github.com/nextcloud/server/blob/055bd5cac381d7173b24408a69b1b9e764092135/apps/dav/lib/Server.php#L162-L167) and therefore [allows access to `/photospublic/` URLs](https://github.com/nextcloud/photos/blob/3f78770243a45e8ada44cb4f4879ea0fd3ed9634/lib/Sabre/PublicAlbumAuthBackend.php#L45).

While it might make sense to explicitly support additional authentication backends in `/public.php/dav` I am not familiar with any of this to know if it would be the right approach or not. Moreover, the DAV client used for public albums [explicitly uses `/remote.php/dav` rather than the default remote URL (`/public.php/dav`)](https://github.com/nextcloud/photos/blob/ef51ed7955ff05093f5163373122807817c05788/src/views/PublicAlbumContent.vue#L134), so for now the provided fix just passes `/remote.php/dav` to `resultToNode` when generating the album node. Note that there are other `resultToNode` calls throughout the code, although I only touched the one that fixes the reproduction steps below (and the added E2E test). It might be necessary to change others as well. 

Another possibility to fix this without explicitly passing the remote URL would have been to remove [the `isPublic` element from the template](https://github.com/nextcloud/photos/blob/903f0daf233583d240957ed7ca4623789f82603c/templates/public.php#L8), so `@nextcloud/files` does not replace `remote.php` with `public.php`. However, conceptually it makes sense to provide the `isPublic` element.

But, moreover, that seems to actually be a brittle hack. Apparently the `isPublic` element [would still be needed in `PhotosAppPublic.vue`](https://github.com/nextcloud/photos/blob/feb0d34869bdd83bd07d33ff3ac88149c0e7725b/src/PhotosAppPublic.vue) (not only based on the comment, but also because if it is removed the public share no longer works); I have not digged into why that one is not used by `@nextcloud/files`, or why the one in the template is not used by whatever uses the one in `PhotosAppPublic.vue`, but it seems that the element from the template is used only during the app startup or something like that :thinking:

Independently of that, note that CI will keep failing until the breaking API change in server is fixed, either by merging https://github.com/nextcloud/photos/pull/3475 or by adjusting the API change in server to be backwards compatible. For now this pull request is just based on #3475 to be able to run the tests.

## How to test

- Create an album
- Add photos to the album
- Create a public share for the album
- In a private window, open the link to the public share
- Open a photo in the viewer
- Download the photo

### Result with this pull request

The photo is download

### Result without this pull request

A 503 Service Unavailable error is returned


artonge (commented):
Looks good, any reason this is still a draft?

danxuliu:
> Looks good, any reason this is still a draft?

I forgot to mention it in the pull request description, sorry :facepalm: I kept it as a draft to ensure that it would not be merged before rebasing first onto master (as it is currently based on #3475, but that is something independent).

Great, and now I undrafted it by mistake with the touchpad :sweat_smile: Anyway, as it has conflicts I will rebase onto master and recompile.

artonge (approved):

susnux (approved):

danxuliu:
Please note that I can not merge myself :-) , as the CI will be red until #3475 is merged or the API break is fixed in server :shrug:

artonge:
What's preventing us to merge https://github.com/nextcloud/photos/pull/3475?

danxuliu:
> What's preventing us to merge #3475?

I kept it as draft because I do not know if the API break will be fixed in server master or not, and if the break is fixed then #3475 is not needed. But... even if it is fixed the change in #3475, although unneeded, would still work. So I have removed the draft state in case you want to merge :-)

PR #3484, opened 2026-05-03
feat: justified grid, density toggle, blurhash crossfade

karlitschek:
Three UI improvements that build on the Vue 3 migration.

1. Consolidate on the justified-row grid algorithm FoldersView was the last surface still using the fixed-square GridLayout component (introduced during Phase 3 to replace vue-virtual-grid). Switch it to TiledLayout, which is the same row-justified algorithm the timeline and album content already use, and delete the now-unused GridLayout/ directory.

   Folders and files are mapped through a small wrapper that adds the `id` and `ratio` fields TiledLayout needs (ratio=1 for both since the folder-listing endpoint doesn't return per-photo dimensions); the inner FileLegacy / FolderComponent rely on `object-fit: cover` to fill whatever tile shape is assigned.

   FileLegacy and FolderComponent also lose the `item.injected.X` indirection inherited from vue-virtual-grid's wrapper shape and read the item directly, matching the upstream Vue 3 migration branch.

2. Blurhash → small → large crossfade in FileComponent The previous code toggled the canvas blurhash and the two img layers via v-if chains driven by `loadedSmall`/`loadedLarge`, producing a hard pop as each layer arrived. Keep all three in the DOM and stack them: blurhash z-index 1, small thumbnail z-2 (200ms fade-in), large preview z-3 (250ms fade-in). Each upper layer covers the one beneath as it becomes opaque, so the visual is a smooth blur → pixelated → sharp progression.

   Bonus fix: beforeUnmount referenced this.$refs.srcLarge (typo) instead of imgLarge, leaving in-flight large-preview loads uncancelled when a tile scrolled out of view.

3. Tile density toggle (small / default / large) New view-density toggle in the timeline header — NcActions menu with three NcActionRadio options. The selection drives a tileBaseHeight computed (mobile 80/120/200, desktop 120/200/320) that's forwarded to FilesListViewer.baseHeight; TiledLayout's row-justification already adapts to any base height.

   Persisted via the existing apps/photos/api/v1/config/{key} endpoint:

   - frontend: new `gridDensity: 'small' | 'medium' | 'large'` field on the userConfig store (default 'medium').
   - backend: UserConfigService.DEFAULT_CONFIGS gets `gridDensity` so it's auto-hydrated into the page initial state alongside the other config keys; ApiController::setUserConfig validates the value is one of small | medium | large.


karlitschek:
@jancborchardt @skjnldsv

susnux (approved):

PR #3485, opened 2026-05-03
feat: search, trip memories, map, slideshow, long-press selection

### LINKED ISSUE #1019: `Locations` link in menu opens in new window
**Describe the bug**
The `Locations` link in the menu opens in a new window as per https://github.com/nextcloud/photos/blob/aec6dfc6676e7dcb18e9b378f2c7b3eef66b6c75/src/router/index.js#L126

Not sure that this is intuitive.

**To Reproduce**
1. Click `Locations`

**Expected behavior**
For it to not open a new window via `_blank` and instead just be the same tab.

### LINKED ISSUE #1090: Action to show Folder on Map if Maps app is enabled
**Is your feature request related to a problem? Please describe.**
Viewing images on maps is nice.

**Describe the solution you'd like**
An action on each folder, which opens the maps app, with the path:
`index.php/apps/maps/m/FOLDERID`

with the nextgovhackathon-master branch you will then see the images of the folder on a map.

**Additional context**
#nextgovhackathon

possible Placement for the button:
image
(just a suggestion, happy for better once)

### LINKED ISSUE #1640: Maps Link seems to be incorrectly generated when nextcloud resides at my.domain/subdirectory
**Describe the bug**
When clicking on *Maps* from the photo app, the address is rewritten to `https://my.domain/apps/maps` where it should be `https://my.domain/nextcloud/apps/maps`

**To Reproduce**
0. Setup nextcloud to reside at a subdirectory of domain
1. Go to *Photos*
2. Click on *Maps*
3. See being redirected to a location without subdirectory appended

**Expected behavior**
Correct redirect

**Additional context**

`'overwritewebroot' => '/nextcloud',`

karlitschek:
Five user-visible additions on top of the Vue 3 migration. Each is intentionally minimum-viable so they ship together; each has clear "deferred to a follow-up" boundaries called out below.

1. Free-text search by filename
   - New PhotosFilter `nameFilter` that emits a DAV `<d:like>` query against `<d:displayname/>`. Special characters in the user's query are escaped (XML + SQL LIKE wildcards) before being embedded in the request body.
   - NcTextField in PhotosApp's `#search` slot, debounced 300ms. Typing pushes a single value into `selectedFilters.name`; the existing filtersQuery / extraFilters pipeline does the rest, so search composes cleanly with the existing date-range and place filters.
   - Deferred: content search ("beach", "sunset") — that requires Recognize-app integration and is out of scope here. The filename + EXIF-anchored search should already feel transformative.

2. Trip memories (/memories)
   - New `services/memories.ts` clusters loaded photos by capture- date gaps (>2 days starts a new trip; clusters of <8 photos are dropped to keep noise out). Cover photo is picked from near the trip's midpoint to avoid arrival/departure shots.
   - New MemoriesView with tile cards (cover + date range + count). Click opens the NC Viewer with the trip's photos as the gallery list.
   - Routed at /memories with a navigation entry next to "On this day". On first visit the view triggers a 500-photo fetch so the page isn't empty before the timeline has been scrolled.
   - Deferred: server-side trip detection (would scale to libraries of millions of photos), and place-anchored / "year ago today" memory cards.

3. Photos map (/map)
   - New MapView using @vue-leaflet's LMap + LMarker, plotting every loaded photo that has GPS metadata. Click a marker → opens NC Viewer with all geotagged photos as the gallery list.
   - `nc:metadata-photos-gps` is now registered in main.ts so the timeline endpoint returns the field; previously it was only registered in sidebar.ts.
   - Centred on the centroid of the first 200 photos so users land somewhere relevant rather than on null island.
   - Replaces the previous /maps route that redirected out to the external Maps app. The inline view works for everyone whether or not Maps is installed.
   - Deferred: leaflet.markercluster integration (real libraries above ~5000 geotagged photos will need it).

4. Photo slideshow
   - New Slideshow component, full-page Teleport, plays through a supplied list of photos with play/pause + prev/next + close and ESC/arrow/space keyboard shortcuts. Auto-advance is 4s.
   - Triggered from a "Slideshow" button in the TimelineView header (visible when no selection is active and at least one photo has been fetched). Plays through the timeline's currently loaded photos in capture order.
   - Deliberately self-contained — the NC Viewer app is in a separate codebase, so adding a slideshow mode there is out-of-scope. This stays a Photos-internal feature.

5. Long-press multi-select
   - FileComponent now starts a 500ms long-press timer on pointerdown; if the press is held that long, the selection toggles and the subsequent click is swallowed. Mouse and touch both go through pointer events, so the same code paths handle desktop right-click-equivalent and mobile tap-and-hold.
   - The existing checkbox-on-hover affordance still works on desktop; long-press is the addition for mobile users for whom the checkbox was a finicky tap target.


karlitschek:
@jancborchardt @skjnldsv

susnux (approved):

PR #3486, opened 2026-05-03
feat: delight pass — favourite pop, selection ripple, shimmer, illustrations, animated counters, album hero, year recap, EXIF overlay

karlitschek:
Eight micro / meso polish items, each at minimum-viable scope. Calls out the deferred bits inline so reviewers know what's deliberately not in scope.

Animated favourite star
- Replace `v-once` on FavoriteIcon (which silently masked any toggle from the bulk-action menu) with a Vue `<Transition>`. The enter animation is a 320ms scale-bounce keyframe with a cubic overshoot timing function; leave fades out in 180ms.
- Skipped: the "particle burst on first favourite of session" bit; that needs session-state plumbing and a one-shot sentinel.

Selection ripple
- The previous `.selected` ring was a hard `outline` glued to the tile edge. Replace with a soft 0.97-scale lift + a 3px primary- colour glow + a 6px-blur drop shadow, all transitioned in 160ms. Keeps the focus ring (different state) for keyboard interaction via `&:focus-within / &:has(:focus)` so focus and selection are visually distinct.

Skeleton shimmer
- Layer a translucent diagonal gradient sweep on top of the blurhash (or the empty primary-element-light background when no blurhash exists) until the small or large preview lands. Pure CSS — translating gradient — so it costs nothing JS-side. Honours `prefers-reduced-motion: reduce`.

Illustrated empty states
- New EmptyIllustration.vue with four hand-drawn-ish SVG compositions (memories / map / faces / timeline). Each uses the Photos accent colour via `var(--color-primary-element)` so it follows the user's theme, including dark mode.
- Wired into MemoriesView and MapView; FacesView / empty-timeline TBD when those surfaces get a polish pass.

Animated counters
- Tiny AnimatedNumber.vue that tween-displays a `value` prop using requestAnimationFrame with an ease-out cubic curve, slot-exposing the live `displayValue` so callers can wrap it in their own translation (e.g. `n('photos', '%n photo', '%n photos', value)`). Honours `prefers-reduced-motion: reduce` by snapping straight to the final value.
- Used in Memories trip cards + the year-recap feature card; can be dropped in elsewhere later without changes.

Album page magazine spread
- New AlbumHero.vue: the album cover photo as a full-bleed 280px- tall hero with the album title overlaid on a darkening gradient. Subtitle combines location (if any) with the photo count via `translatePlural`. Hooks a passive scroll listener to translate the cover background up to 60px slower than the page for a soft parallax effect.
- Skipped (called out in commit so they're easy to find): album cover picker UI and accent-colour extraction from the cover image. Both worthwhile follow-ups; both larger than this round.

Year-in-review recap
- New `buildYearRecap` in services/memories.ts: groups loaded photos by year, picks the most-recent year with at least 30 photos, then curates a target-of-60 set by including all favourites first then sampling evenly across the calendar so December doesn't dominate. Cover is the curated-set midpoint (avoiding the "first or last photo of the year" trap).
- Surface as a hero feature card on MemoriesView (above the trip grid) with eyebrow / title / count metadata. Click opens the curated set in the existing in-app Slideshow component, so the flow is "click → autoplay slideshow → close to return" with no cross-app coordination needed.
- Skipped: text-card sections between groups, music, and "X years ago today" cards. They're separate algorithms.

EXIF overlay in the slideshow
- Press `i` in the slideshow to toggle a frosted-glass aside that surfaces camera (Make + Model from IFD0), aperture, focal length, exposure, and ISO. Renders nothing when no EXIF data is available so we don't show an empty panel.
- Required registering `nc:metadata-photos-exif` and `nc:metadata-photos-ifd0` in main.ts — previously these were only fetched by the file sidebar.


karlitschek:

susnux (approved):

PR #3487, opened 2026-05-03
feat: per-photo actions menu (EXIF / Add to album / Share / Delete)

karlitschek:
Adds a 3-dot overflow menu on every photo tile so EXIF / Add to album / Share / Delete are reachable without first entering selection mode. Stacked on top of #3486.

PhotoActionsMenu.vue (new) — NcActions popup with 4 entries; "View metadata" opens an inline NcDialog reusing the same field-selection logic as the slideshow overlay; Delete prompts confirmation before emitting.
FileComponent.vue mounts the menu next to the existing favourite/select affordances; opacity-0 by default, revealed on :hover / :focus-within. showActionsMenu prop (default true) lets picker contexts opt out.
TimelineView.vue wires the three request-* events: Add to album re-uses AlbumPicker via singleFileForAlbumPicker; Share navigates to /apps/files/files/{fileid}?opendetails=true (NC34 removed the legacy OCA.Files.Sidebar.open(path) and getSidebar() needs Files-app context Photos doesn't have); Delete dispatches deleteFiles after optimistically dropping the id.
PhotosPicker.vue passes :showActionsMenu="false" so the picker stays focused on picking.
Test plan
 Hover a tile → 3-dot button fades in
 View metadata dialog shows camera + aperture + focal length + exposure + ISO + filename
 View metadata empty-state renders for photos without EXIF
 Add to album: AlbumPicker adds only that photo, not the bulk selection
 Share: navigates to Files app on the photo with sidebar open
 Delete: confirm dialog → tile vanishes; cancel keeps it
 PhotosPicker tiles have no 3-dot menu
 Clicks on the menu don't open the slideshow viewer


karlitschek:

susnux (approved):

PR #3488, opened 2026-05-03
Frank/feat/photos perf index

karlitschek:
Backend

PhotoIndexMapper::getEnrichedTimelineForUser does a single SQL: photos_index ⨝ filecache ⨝ vcategory_to_object ⨝ vcategory — returns rows with path, etag, size, mtime, taken_at, permissions, favorite per file. EXIF/GPS/IFD0/blurhash/dimensions come from one batched IFilesMetadataManager::getMetadataForFiles call — same cache the metadata pipeline already uses.
Query is scoped to the user's primary storage so DAV source URLs map cleanly. Group folders / external storage stay on the DAV fallback for now.
Frontend

IndexedTimelineSearch.ts hits /api/v1/index/timeline and produces File objects with the same attribute shape the legacy DAV fetcher produces — downstream consumers (FilesByMonthMixin, FileComponent, Slideshow, album-add) don't change.
FetchFilesMixin now routes through the indexed path when indexStatus.ready is true. Falls back to DAV on any error. onThisDay / onlyFavorites / extraFilters queries always use DAV — those filter shapes aren't implemented in the indexed endpoint yet.
Caveats to watch when testing:

First load on a fresh install: ready=false → still uses DAV, banner visible. After occ background-job:execute OCA\\Photos\\Jobs\\PhotoIndexBackfillJob (or waiting for cron), ready=true → switches to indexed.
Group-folder photos: indexed path filters them out by design. They'll reappear when the next PR adds per-user DAV-path mapping at index time.
Filter UI (date / place / tag): falls through to DAV unchanged.


karlitschek:

PR #3489, opened 2026-05-03
Frank/feat/photos search indexed

karlitschek:
feat: indexed-API search bypasses broken DAV SEARCH

Stacks on top of #3488. NC34 dev builds throw inside FilesMetadataManager::getKnownMetadata() during DAV SEARCH backend setup (strict lazy-AppConfig validation: "The loading of lazy AppConfig values have been triggered by app 'core'") — every DAV SEARCH request 500s, which surfaces as "search is broken". Sidesteps the problem by serving search out of our own backend against oc_photos_index.

Backend: PhotoIndexMapper::searchUserTimeline joins photos_index ⨝ filecache ⨝ favorites ⨝ systemtag_object_mapping ⨝ systemtag with (filecache.name LIKE %q% OR systemtag.name LIKE %q% OR taken_at IN dateRange). Date grammar is tight (year, year-month) so generic words don't accidentally match. IndexController::composeItem is shared between timeline + search responses.

Client: new getIndexedSearchPhotos(query, options). FetchFilesMixin.fetchFiles is a three-way: search + ready → /api/v1/index/search; no search + ready + no exotic filters → /api/v1/index/timeline; else DAV. Falls back to DAV silently on HTTP failure. The legacy DAV nameFilter was also broadened to match name/tag/date so instances on the DAV-only path get the same behaviour.


karlitschek:
@jancborchardt @AndyScherzinger @artonge  @skjnldsv

artonge (changes requested):
Overall, I don't think we want an index.

If the issue is with the performance of the SEARCH WebDAV endpoint that we are using, then we should fix it on the server side. FYI, it already improved by a lot recently. There is also a lot of room for improvement on the frontend side of Photos to make smarter search request with a lower and higher date limit.

If the issue is with the features that the WebDAV endpoints supports, then maybe we can come up with a more lightweight targeted solution.

The big plus of using the WebDAV API:
1. It is a good dogfooding of our WebDAV API, performance and feature set
2. It simplifies the complexity of the app by a lot, both backend and frontend wise
3. It does not require an initialisation step

PR #3490, opened 2026-05-03
chore: bump app version to 7.0.0-dev.1 and fix the db migration

karlitschek:
Force NC's `app:upgrade` to pick up the new schema migrations and background-job registrations from the perf-index / search-indexed / hls-transcoding branches. The version bump is the trigger that makes `occ app:enable photos` (and `occ upgrade`) re-run the migration loader for this app.


susnux:
Done alredy :white_check_mark:

PR #3496, opened 2026-05-04
feat(scrubber): modern glassy redesign with density tick marks

### LINKED ISSUE #426: Showing dates/years on timeline scrollbar

### LINKED ISSUE #2713: Impossible to navigate
I haven't looked at my NextCloud server in months, but today I needed to find a photo from a specific date range.

This is apparently not possible. In fact, there appear to be *no* methods of sorting, filtering, or even *navigating* my photo collection apart from scrolling with the mouse wheel through thousands of photos. The web interface doesn't even have a scroll bar anymore.

The default sort appears to be reverse chronological, but starting at June 2018. There is no way to see the most recent photos or change this behavior. There are no settings of any kind, no controls, nothing. I have no idea how many photos there are because the scroll bar is gone, there's no way to find the most recent photos.

This is quite simply unusable. This app has zero utility as a photo browser. It would take far less time and effort to download my entire 500GB nextcloud archive, unzip it, and manually locate the file I want in my file browser. Windows Explorer is a better photos app than this, and that's really saying something.

karlitschek:
The right-edge date scrubber gets a state-of-the-art makeover:

Visual direction
- Idle: a hairline gradient strip that just hints "drag here". No tick marks, no labels, nothing competing with the photo grid.
- Hover / scrub: the strip glassens, year pills slide in from the right, tick marks light up across the track, and the thumb expands into a glassy capsule that shows the active month inline.
- The previous floating-tooltip-next-to-the-thumb is gone; the date now lives *inside* the thumb pill, so the user's eye doesn't have to dart between two surfaces while scrubbing.

Density tick marks (the new affordance)
- One tick per month, width proportional to that month's photo count. Reads as a tiny histogram of where the user's library is dense — January's vacation snaps stand out as a bigger tick next to a quiet February.
- Width derived from a `--tick-density` CSS custom property bound at the element so the renderer keeps everything visual (no per-tick inline-width recompute on resize).
- 95th-percentile clamp on the max so a single bursty month doesn't crush the scale and turn every other month into a flat short tick.
- Empty months still render at the minimum 4px so the rhythm stays consistent.

Other touches
- Year pills are pill-shaped + frosted-glass (NC `color-mix` + `backdrop-filter`); the year that contains the active month gets a stronger treatment so the user knows where the thumb sits at a glance.
- Thumb expansion uses an ease-out-quint curve (cubic-bezier 0.22, 1, 0.36, 1) for the same "deliberate" feel the photo hover effect uses.
- Active scrubbing adds a soft halo (8px primary-element shadow at 30% mix) so the user feels they have a firm grip.

Wiring
- TimelineView exposes `monthCounts` (built from `fileIdsByMonthUngrouped` so the histogram reflects the true distribution, not the visible-tile count after burst stacking).
- DateScrubber takes a new optional `monthCounts` prop. Falls back gracefully when absent (uniform-rhythm strip, same drag UX).

Build refresh attached.


karlitschek:

jancborchardt (commented):
Will check :) 
@karlitschek as with the other pull requests as well, it would be very helpful to have before/after screenshots, to quickly know what to look out for.

skjnldsv:
And I see a lot of them are based onto other branches, which makes it very hard to know what the actual set of changes is this about.

Could you adjust and make each PR against master ? :)
Or maybe explain the order we should review if they are all dependent on each others ?

susnux (approved):

susnux:
Rebased onto master as a single feature as per PR description.
Applied to Vue 2 master as Vue 3 migration is currently blocked (hope to resolve that this week).

PR #3498, opened 2026-05-04
Drop album cogwheel icon on filtered albums

karlitschek:
Removes the CogOutline overlay shown on album cards in the Albums overview when an album has filters set.


skjnldsv:
cc @nextcloud/designers, is it something we want ? I have no insight on the matter :)

susnux (commented):
If we want this then the code changes look ok (simple remove).
So purely a design decision.

jancborchardt (approved):
Looks good, is not necessary to indicate the filter. 👍

skjnldsv (approved):

susnux (approved):

PR #3553, opened 2026-06-05
feat: display video duration

### LINKED ISSUE #3472: Low contrast for camera icon on video thumbnails
**Describe the bug**
While looking at the list of media, there is a camera icon on the thumbnail for video files. This icon has low contrast and if the video thumbnail has dark colors, then it's hard (or impossible) to see. This makes it hard to quickly distinguish video files from photos.

**To Reproduce**
1. Upload a video file that will have a dark thumbnail.
2. Open photos.
3. See that the camera icon is barely visible.

**Expected behavior**
I expect the camera icon to have contrasting colors to the thumbnail, or an outline that would give it contrast regardless of the thumbnail.

**Screenshots**

Image

 - OS: Linux
 - Browser Firefox
 - Version Any

 - Device: NA
 - OS: NA
 - Browser NA
 - Version NA

**Browser log**
```
Open your console, reload your page and/or do the action leading to this issue and copy/paste the log in this thread.
```

**Additional context**
-

luka-nextcloud:
Resolves: #3472

before:
image

after:
image


src/utils/fileUtils.ts
```diff
@@ -171,3 +171,32 @@ export function legacyToViewerFileInfo(file: FoldersNode) {
 		permissions: file.permissions,
 	}
 }
+
+export async function getVideoDurationFromUrl(url) {
```
artonge: Could this be extracted and stored with the file metadata API by the server? Similar to what we do with EXIFs: https://github.com/nextcloud/photos/blob/30bdf8eb99659a09c215d0432c40893189e21f58/lib/Listener/ExifMetadataProvider.php

luka-nextcloud: @artonge As far as I know, CLI tool `ffprobe` is required to extract the video duration time on the server side. Since this tool is not installed by default for all servers, I don't think we can go with this approach.

src/utils/fileUtils.ts
```diff
@@ -171,3 +171,32 @@ export function legacyToViewerFileInfo(file: FoldersNode) {
...
+	return new Promise((resolve, reject) => {
+		const video = document.createElement('video')
+		video.preload = 'metadata'
+
+		const cleanup = () => {
+			video.removeAttribute('src')
+			video.load()
```
artonge: Why do we need to call `load()` here?

src/utils/fileUtils.ts
```diff
@@ -171,3 +171,32 @@ export function legacyToViewerFileInfo(file: FoldersNode) {
...
+		}
+
+		video.onloadedmetadata = () => {
+			const seconds = video.duration
+			cleanup()
+			if (!Number.isFinite(seconds)) {
+				reject(new Error('Could not read finite duration.'))
```
artonge:
```suggestion
				reject(new Error('Could not read non-finite duration.'))
```

src/components/FileComponent.vue
```diff
@@ -243,6 +252,23 @@ export default {
 			ctx.putImageData(imageData, 0, 0)
 		},
 
+		async getVideoDuration() {
+			if (!this.isVideo) {
+				return
+			}
+
+			const totalSeconds = parseInt(await getVideoDurationFromUrl(this.file.source) as string)
```
artonge: Can we make `getVideoDurationFromUrl` to return an number instead?

src/components/FileComponent.vue
```diff
@@ -16,7 +16,10 @@
 
 			<!-- image and loading placeholder -->
 			<div class="file__images">
-				<VideoOutline v-if="file.mime?.includes('video')" class="icon-overlay" :size="64" />
+				<div v-if="isVideo && videoDuration" class="duration-time" aria-hidden="true">
```
artonge: I would display the icon even if we do not have the duration

artonge: Also, why `aria-hidden="true"`?

artonge:
```suggestion
				<div v-if="isVideo && videoDuration" class="file__duration" aria-hidden="true">
```

artonge (commented):
Can you post before/after screenshots?

src/utils/fileUtils.ts
artonge: Alright. I fear the performance limitation of such solution, but let's go with it for now, we can always improve later on.

src/components/FileComponent.vue
```diff
@@ -16,7 +16,10 @@
 
 			<!-- image and loading placeholder -->
 			<div class="file__images">
-				<VideoOutline v-if="file.mime?.includes('video')" class="icon-overlay" :size="64" />
```
artonge, line 19: Can you remove the `icon-overlay` css?

artonge (approved):

PR #3571, opened 2026-06-21
Fix timestamp when looking at photos

### LINKED ISSUE #2768: New photos or videos not displayed in "All Media"; new Photos not displayed in "Photos"; new videos ARE displayed in "Videos"
**Describe the bug**
From the Nextcloud Android client, I auto-upload photos and videos into the /Photos directory of my Nextcloud instance.  Uploaded photos or videos are not displayed in the "All Media" view and photos are not displayed in the "Photos" view.  Oddly, new videos ARE displayed in the "Videos" view.  

All uploaded photos and videos are displayed in the "Folder" view.

Also, all photos and videos from files_external mounted into "/Photos/Shared" are displayed in all of the views.

There are no errors in the browser log, or the PostgreSQL server; running Nextcloud 30.0.5. Same result with Firefox or Chrome.

tintou:
Reject timestamp 0 as it is very unlikely to represent the right time and least to photos being unsorted.

This is especially true for my setup using the S3 from OVH that seem to not return the mtime.

## 🤖 AI (if applicable)

- [X] The content of this PR was partly or fully generated using AI

Fixes: https://github.com/nextcloud/photos/issues/2768

### DIFF
--- a/lib/Listener/OriginalDateTimeMetadataProvider.php
+++ b/lib/Listener/OriginalDateTimeMetadataProvider.php
@@ -38,7 +38,7 @@ class OriginalDateTimeMetadataProvider implements IEventListener {
 		try {
 			// Note: We do not have the timezone when parsing the date, so the timestamp will be off by X hours.
 			$dateTime = DateTime::createFromFormat($format, $date);
-			if ($dateTime !== false) {
+			if ($dateTime !== false && $dateTime->getTimestamp() > 0) {
 				return $dateTime->getTimestamp();
 			}
 			return false;
--- a/lib/Listener/OriginalDateTimeMetadataProvider.php
+++ b/lib/Listener/OriginalDateTimeMetadataProvider.php
@@ -112,6 +112,9 @@ class OriginalDateTimeMetadataProvider implements IEventListener {
 		}
 
 		// Fallback to the mtime.
-		$metadata->setInt(self::METADATA_KEY, $node->getMTime(), true);
+		$mtime = $node->getMTime();
+		if ($mtime > 0) {
+			$metadata->setInt(self::METADATA_KEY, $mtime, true);
+		}
 	}
 }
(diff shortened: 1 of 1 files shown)


lib/Listener/OriginalDateTimeMetadataProvider.php
```diff
@@ -67,13 +67,6 @@ public function handle(Event $event): void {
 			return;
 		}
 
-		// We need the file content to extract the EXIF data.
-		// This can be slow for remote storage, so we do it in a background job.
-		if (!$node->getStorage()->isLocal() && $event instanceof MetadataLiveEvent) {
-			$event->requestBackgroundJob();
-			return;
-		}
-
```
come-nc, line 77: I’m not sure I understand the rationale.
The code below reads exif metadata, that does access the file bytes, so the guard was correct and the commit comment is wrong?

tintou: It is not reading the exif but getting the cached exif here (from what I understand from the code)

susnux: how is this related to the problem you want to fix?

tintou: It helped in my specific case but that's not part of the issue you're right, removed the commit

lib/Listener/OriginalDateTimeMetadataProvider.php
```diff
@@ -112,6 +112,9 @@ public function handle(Event $event): void {
 		}
 
 		// Fallback to the mtime.
-		$metadata->setInt(self::METADATA_KEY, $node->getMTime(), true);
+		$mtime = $node->getMTime();
+		if ($mtime > 0) {
+			$metadata->setInt(self::METADATA_KEY, $mtime, true);
+		}
```
artonge, line 119: Let's keep meaningful fallback, else, the file won't even be included in requests filtering by `photos-original_date_time`.

```suggestion
		$mtime = $node->getMTime();
		if ($mtime > 0) {
			$metadata->setInt(self::METADATA_KEY, $mtime, true);
			return;
		}
		
		$metadata->setInt(self::METADATA_KEY,  $node->getUploadTime(), true);		
```

PR #3760, opened 2026-08-16
fix(album): roll back failed database transactions

joshtrichards:
Ensure album database transactions are rolled back when an exception occurs during multi-step operations.

Prevents partial collaborator updates or album deletions when a later database operation fails.

- Roll back the transaction and rethrow the original exception when `setCollaborators()` fails.
- Apply the same transaction-safety pattern to `delete()`.
- Catch `\Throwable` so both exceptions and PHP errors cannot leave a transaction open.
- Preserve the existing commit behavior when all operations complete successfully.

Since this was a tiny fix didn't think bother to use the `TTransactional` helper.

## 🤖 AI (if applicable)

- [ ] The content of this PR was partly or fully generated using AI

### DIFF
--- a/lib/Album/AlbumMapper.php
+++ b/lib/Album/AlbumMapper.php
@@ -164,10 +164,10 @@ class AlbumMapper {
 			$this->connection->commit();
 		} catch (\Throwable $e) {
 			$this->connection->rollBack();
- 			throw $e;
- 		}
- 	}
-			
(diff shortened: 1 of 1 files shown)


susnux (approved):

susnux:
Linting needs to be fixed

artonge (approved):

PR #3791, opened 2026-08-23
feat: zoom and pan photos in the slideshow

karlitschek:
## What

The Photos **slideshow** (`PhotoSlideshow.vue`) could only show each photo fitted to the screen — no way to look closer. This adds zoom and pan:

- **Pinch** to zoom + **drag** to pan on touch
- **Mouse wheel** to zoom + **drag** to pan with a pointer
- **Double-click / double-tap** to toggle between fit and zoom
- **`+` / `-` / `0`** to zoom in / out / reset from the keyboard, so pointer gestures aren't the only way in (accessibility)

## Behaviour details

- **Pan is clamped** so a photo can never be dragged off screen, and **zoom resets when the shown photo changes** (manual navigation or the slideshow advancing).
- **Zooming pauses the slideshow** so it doesn't advance past the photo you're inspecting.
- **Swipe still works:** while a photo is at fit (1×), a horizontal drag reaches the modal as a swipe to the next photo; once zoomed, drags pan instead and are kept from the modal's swipe navigation.
- Motion honours **`prefers-reduced-motion`**, and the zoom transition is dropped mid-gesture so a pinch/drag tracks the fingers exactly.

## Structure & tests

- The pan/zoom maths (scale clamping, pan-offset clamping, pinch distance, double-tap toggle) lives in a small pure `src/utils/panZoom.ts` with a full `panZoom.spec.ts` (13 cases). The component just wires pointer/wheel/keyboard events to it.
- Full `npm test` green (78 tests). `npm run lint` clean for the changed files; the pre-existing repo-wide `ts:check` errors are in unrelated files (this change adds none).

## Verification

Verified on a dev instance: keyboard `+`/`-` zoom (transform goes 1× → 2×), double-click toggles back to fit, and the zoom persists rather than being reset by auto-advance (confirming the autoplay pause). Screenshot of a photo zoomed inside the slideshow attached in review.

Source only; `js/`/`css/` will need the usual asset compile.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/components/PhotoSlideshow.vue
+++ b/src/components/PhotoSlideshow.vue
@@ -16,11 +16,28 @@
 		@previous="showPrevious"
 		@next="showNext"
 		@close="emit('close')">
-		<img
+		<!-- Zoom/pan surface: pinch or wheel to zoom, drag to pan, double-tap to toggle. -->
+		<div
 			v-if="currentPhoto !== undefined"
-			class="slideshow__photo"
-			:src="getPreviewUrl(currentPhoto, PREVIEW_SIZE)"
-			:alt="currentPhoto.basename">
+			ref="viewport"
+			class="slideshow__viewport"
+			:class="{
+				'slideshow__viewport--zoomed': scale > MIN_SCALE,
+				'slideshow__viewport--interacting': interacting,
+			}"
+			@pointerdown="onPointerDown"
+			@pointermove="onPointerMove"
+			@pointerup="onPointerUp"
+			@pointercancel="onPointerUp"
+			@wheel.prevent="onWheel"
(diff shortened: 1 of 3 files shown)


susnux (commented):
I do not think we should put more effort into this, instead we should directly improve the viewer instead to also benefit from the improvements in other apps.
It also adds another maintenance burden as we duplicate the exact same feature in both apps

karlitschek:
> I do not think we should put more effort into this, instead we should directly improve the viewer instead to also benefit from the improvements in other apps. It also adds another maintenance burden as we duplicate the exact same feature in both apps

True! I agree.

Issue #425, opened 2020-08-23
five star rating of pictures


wein-geist:
**How I do it now**
Assume having a folder containing the pictures of an event/vacation/etc. I usually create a subfolder "best-of" or similar containing a subset with the best pictures. This leads to duplicates of pictures and is kind of a messy way to organize photos.

**Describe the solution you'd like**
I would like to suggest a rating system for Photos. Every user could develop his own rules. Mine would be for example 
- 1*: default
- 3*: best 20 pics
- 4*: best 5 pics
- 5*: absolute masterpiece, maybe 20 pictures in all of my photos. 

In order to do this, following features are needed:
- initially, all pictures are unrated
- bulk rating per album/folder should be possible (eg. set all to ***..)
- Set rating directly when viewing them, therefore the viewer would need to feature a rating element.
- When viewing an album/folder, one should be able to filter e.g. tree stars and above
- Cherry on top: when sharing an album, only share 3 stars or better

**Describe alternatives you've considered**
Folder with a subset copy of the best 20 something pictures.
At the moment there is a star feature in Photos. This doesn't allow for a more defined rating, plus you cant filter albums by starred pictures, only a global view of them.

skjnldsv:
@jancborchardt

alexanderdd:
I think we need a decision if ratings will be implemented or not. How do we make that decision?
If yes, we should open separate reports for individual steps that you named after "following features are needed:"

In your idea, is the rating also saved in the file in metadata?
Is the rating of a foto the same for all users?
 (I think that both would be good)

jancborchardt:
There should be the ability to favorite photos of course, like in Files and other apps of Nextcloud. It should also of course be symmetric to the favoriting in Files, reflecting the status the same in each app.
Currently it’s not very easy to favorite a photo largely because the menu is missing in the viewer: https://github.com/nextcloud/viewer/pull/15

> you cant filter albums by starred pictures, only a global view of them.

That is a very good point @dlauper, and we should fix that by sorting favorites inside a folder up top of that folder so they are directly visible. Just like we do in Files, Notes, Talk and other apps.

-------

On the point of a rating scale: A full rating scale from 1 to 5 is out of scope here and much too complex for the majority of users. We’re not aiming to be Adobe Lightroom. ;)

Also comparing to other related solutions:
- Google Photos has simple on/off favoriting https://www.theverge.com/2018/5/21/17377738/google-photos-favorites-new-feature
[...]

PR #3793, opened 2026-08-25
feat: add captions, descriptions and star ratings

karlitschek:
## What

Adds user-editable **descriptions/captions** and **star ratings** to photos. Nextcloud stores neither in an editable way today (the metadata index only exposes date/time and GPS as writable), so this adds a small store for them in the Photos app and surfaces it in the existing metadata dialogs.

- **Edit metadata** dialog gains a description field and a 0–5 star rating.
- **View metadata** dialog shows the description and the rating.

## Storage & API (backend)

- A new `photos_metadata` table holds `{ file_id, description, rating }`, one row per file (migration + `PhotoMetadata` entity + `PhotoMetadataMapper`).
- A `MetadataController` exposes `GET`/`PUT /api/v1/metadata/{fileId}`. **Access is always scoped to the current user's own files:** a file id is only ever resolved through `getUserFolder()->getById()`, so a caller can neither read nor write details for a file they cannot see (→ 404), and writing additionally requires `PERMISSION_UPDATE` on the file (→ 403). CSRF protection is kept on the write; ratings are validated to 0–5.
- Metadata is stored in the app's DB (like albums), **not** written into the image/EXIF — so it doesn't travel with the file outside Nextcloud. This was a deliberate choice over server-side EXIF/XMP writing (which would need exiftool and modify user files); noted here as the main trade-off.

## Frontend

- A small `photoDetails` service (axios) and a reusable `StarRating` component (uses `--color-favorite`, so it stays legible in dark mode).
- The two dialogs fetch the details alongside the EXIF they already load, and the edit dialog saves them next to the date/GPS update.

## Verification

`composer psalm` and `composer cs:check` are clean; `php -l` passes; `npm run lint`, `npm run build`, and `npm test` (65) are clean for the changed files (the pre-existing repo-wide `ts:check` errors are in unrelated files). Verified end-to-end on a dev instance: the API round-trips and rejects an out-of-range rating (400); editing a photo's description + rating persists and shows up in **View metadata** — no console errors.

Source only; `js`/`css` will need the usual asset compile.

### Possible follow-ups
- Clean up the row on file deletion (a `NodeDeletedEvent` listener) to avoid orphan rows.
- A PHPUnit test for the controller's permission/validation logic.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/appinfo/routes.php
+++ b/appinfo/routes.php
@@ -14,6 +14,18 @@ return [
 			'verb' => 'PUT',
 		],
 		[
+			'name' => 'metadata#get',
+			'url' => '/api/v1/metadata/{fileId}',
+			'verb' => 'GET',
+			'requirements' => ['fileId' => '\d+'],
+		],
+		[
+			'name' => 'metadata#update',
+			'url' => '/api/v1/metadata/{fileId}',
+			'verb' => 'PUT',
+			'requirements' => ['fileId' => '\d+'],
+		],
+		[
 			'name' => 'api#serviceWorker',
 			'url' => '/service-worker.js',
 			'verb' => 'GET',
--- a/lib/Controller/MetadataController.php
+++ b/lib/Controller/MetadataController.php
@@ -1,0 +1,109 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * SPDX-FileCopyrightText: 2026 Nextcloud GmbH and Nextcloud contributors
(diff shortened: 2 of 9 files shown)


susnux:
@jancborchardt did we change our opinion on #425 ?
Because from the discussion there star rating was nothing we wanted if I understood it correctly.

PR #3832, opened 2026-09-03
fix(media-folders): do not re-create deleted source folders

silverkszlo:
### Bug
Opening the Photos app fetches media across all configured source folders in one request. If a source folder had been deleted, that request 404'd, and Photos silently recreated the folder (empty, without its content) as a side effect of handling that error. So simply navigating to Photos after deleting a folder brought it back (empty).

Restoring the original folder from Files' trash then produced a duplicate: the original path was already occupied by the empty recreated folder, so the restored one landed with a "(restored)" suffix instead, holding the original content.

### Step to reproduce
1. In Files app create a folder and add some media files in it
2. Go to Photos app and navigate to Photos settings
3. Click on "Add folder" and select to the newly created folder
4. Verify the All media dashboard shows the media files that this folder contains
5. Now "Delete" the folder in the Files app and navigate to Photos again - the files should not be present
6. Navigate back to Files and observe that the folder has been recreated without content
7. Navigate to "Deleted files" in Files app and restore the deleted folder
8. The folder with content is restored but with the suffix "restored"

### Solution
- Only the app's own upload folder is still auto-created when missing; a user-added source folder no longer is
- Fix the 404-detection that gated this: the match pattern had an extra slash, and it read a `.data` property that doesn't exist on the dav client's error object, so the check never ran
- A missing source folder now shows an error inviting its removal from settings, instead of silently recreating it

This stopped Photos from leaving an empty placeholder behind after a source folder was deleted, which caused Files' trash restore to rename the restored folder with a "(restored)" suffix due to the path conflict.

## 🤖 AI (if applicable)

- [x] The content of this PR was partly or fully generated using AI


src/components/Settings/PhotosSourceLocationsSettings.vue
```diff
@@ -18,7 +18,7 @@
 				:key="index">
 				<PhotosFolder
 					:path="source"
-					:can-delete="photosSourceFolders.length !== 1"
+					:can-delete="true"
```
susnux:
```suggestion
					can-delete
```

src/mixins/FetchFilesMixin.ts
```diff
@@ -90,20 +91,30 @@ export default defineComponent({
...
+								this.resetFetchFilesState()
+								return []
+							} catch (error) {
+								this.errorFetchingFiles = 404
+								logger.error('Fail to create source directory', { error })
+							}
+						} else {
 							this.errorFetchingFiles = 404
-							logger.error('Fail to create source directory', { error })
+							logger.error(`The ${missingFolder} media folder does not exist.`)
```
susnux: please keep the error in the log 


```suggestion
							logger.error(`The ${missingFolder} media folder does not exist.`, { error })
```

src/mixins/FetchFilesMixin.ts
```diff
@@ -90,20 +91,30 @@ export default defineComponent({
...
-						logger.debug(`The ${source} folder does not exist, creating it.`)
-						try {
-							await davClient.createDirectory(join(defaultRootPath, source))
-							this.resetFetchFilesState()
-							return []
-						} catch (error) {
+					const { photosLocation, photosSourceFolders } = store.state.userConfig
+					const errorBody = await getErrorBody(error)
+					const missingFolder = photosSourceFolders
+						.find((source) => errorBody.includes(`File with name ${source} could not be located`))
```
susnux, line 97: intentionally dropping the leading / ?

susnux (commented):
Makes sense, just some nit-picks.
But still need to test this.

Could you maybe ask you LLM to also add a playwright test for this?

src/mixins/FetchFilesMixin.ts
silverkszlo: yes because the source/entry in `photosSourceFolders` already comes with a leading slash

susnux (approved):

PR #3174, opened 2025-09-10
chore: Migrate to vue3

artonge:
- Update dependencies for Vue3
- Run `npm lint:fix` and fix reported warnings
- Update usage of `vue`, `vue-router` and `vuex`
- Remove unused `isMobile` import
- Get rid of `vue-virtual-grid` Replaced with the internal virtual scrolling component. 
- Modernize router instantiation
- Do not use Vue method to update objects
- Use `vue` provided `nextTick` function
- Mark semaphore object as raw
- Migrate to Vue3 lib of leaflet 

## TODO

- [ ] Migrate `UploadPicker`@nc/vue

PR #3855, opened 2026-09-11
refactor: move the components to script setup

skjnldsv:
Stacked on #3876, review the last six commits only. What is left of the original
#3855 once the Pinia move (#3874), the named exports (#3875) and the Vue 3 switch
(#3876) are their own PRs.

The mixins become composables in `src/composables/`, then the components move
to `<script setup lang="ts">` one area per commit: grid, albums,
faces/tags/places, timeline and settings. The last commit deletes
`src/mixins`.

Relates to #3174

This comes after the flip because the components use `useTemplateRef`,
`defineSlots`, `useId` and the tuple `defineEmits`, none of which exist on
Vue 2.7.

Lint, vitest and the build pass, `vue-tsc` goes from 108 errors to 25, the
rest is in files master already fails on plus `SeparatorVirtualGrid`, which
nothing renders. Playwright, full suite locally: green.


package.json
```diff
@@ -45,7 +45,7 @@
     "@mdi/svg": "^7.4.47",
     "@nextcloud/auth": "^2.1.0",
     "@nextcloud/axios": "^2.1.0",
-    "@nextcloud/dialogs": "^7.4.1",
+    "@nextcloud/dialogs": "~7.4.1",
```
susnux, line 48: Why?

susnux (commented):
Honestly this is unreviewable, this will take a good amount of review work due to the size and changes directly undone by following changes.

Please keep such PRs - especially when AI assisted - condensed.
Meaning especially with AI its easy to group changes to one set and split it into multiple PRs for easier reviewing.

It would be much easier to review to split this into a small number of individual PRs (and e.g. use github stacked PRs):
- Migrate to pinia (can be done with Vue 2 as a preparation)
- Migrate from mixins to composables
- Migrate to script setup (also doable in Vue 2)
- Update linting (e.g. the sass changes) and resolve its problems reported (still no Vue 3 needed - could be done in any order)
- Migrate to Vue 3

package.json
skjnldsv: Asked AI to resolve a dep conflict... I shouldn't have 🤣🤣
I need to do a final check up anyway, there's more to double check

skjnldsv:
@susnux this is not ready for review 😭😭
But yes, I completely agree, this will need to be split 👍👍

PR #3874, opened 2026-09-15
refactor(store): move from Vuex to Pinia

skjnldsv:
First of four PRs that split up #3855. This one only moves the state: the
Vuex modules become Pinia setup stores, `vuex` and `vuex-router-sync` go, the
app stays on Vue 2.7 and the components keep their Options API.

Relates to #3174

`Vue.set` and `Vue.delete` stay in the stores for now, Pinia runs on the Vue 2
reactivity here. They become plain assignment and `delete` with the Vue 3
switch.

`@nextcloud/dialogs` is pinned to `~7.4.1` until then: 7.5.0 peers Vue 3.

Typing the stores surfaced a few call sites the untyped `dispatch` hid, fixed in
the commits that touch them (a `downloadFiles` action no module defined, a
`removeTag` looking names up under the wrong key, a 404 stored as a
collection). The commit messages have the details.

Lint, vitest (111 tests, the stores are now covered) and the build pass.
`vue-tsc` goes from 364 errors on master to 347. Playwright, full suite
locally: green.


package.json
```diff
@@ -45,7 +45,7 @@
     "@mdi/svg": "^7.4.47",
     "@nextcloud/auth": "^2.1.0",
     "@nextcloud/axios": "^2.1.0",
-    "@nextcloud/dialogs": "^7.4.1",
+    "@nextcloud/dialogs": "~7.4.1",
```
susnux: ?

susnux (approved):
Also contains some fixes it seems (sanitize true / false to 1 / 0 for favorite state in webdav)

susnux:
Well Claude has a different opinion than me:

> In FaceContent.vue, onPhotoDeleted should call removeFileIdsFromFace, not removeFilesFromFace.
> 
> Vuex had a mutation and an action under the same name — the old code used the mutation, so it only updated local state. The new call hits the action, which deletes from the server. By that point deletePhoto has already removed the photo from the files store, so the action looks up files[fileId], gets undefined, and throws on file.basename. Deleting a photo from a person's page fails.
> 
> Every other converted commit in the PR picked the state-only variant, so this one looks like a slip.
>

marcelklehr (approved):
Looks good, apart from Claude's finding

skjnldsv:
Fixed both comments, let's merge

PR #3876, opened 2026-09-15
feat: switch to Vue 3

skjnldsv:
Stack: #3879 → **#3876** → #3877

Vue 3.5, vue-router 5, pinia 4, `@nextcloud/vue` 9, `@nextcloud/vite-config` 2, `@vue-leaflet/vue-leaflet` in place of `vue2-leaflet`. The components keep the Options API, the move to `<script setup>` is #3877, and the templates were already rewritten in #3879, so this diff only holds what Vue 3 and the new library majors required.

Relates to #3174

`@nextcloud/upload` is dropped, its latest still peers Vue 2. The picker is `NcUploadPicker` from `@nextcloud/vue` and the uploader comes from `@nextcloud/files/upload`, hence `@nextcloud/files` on the 4.1 beta.

Two things Playwright caught that the build did not: `:param*` is a repeatable param in vue-router 4+, so `/albums` matched the album route with an empty array and every collection request went to `albums//`; and `SemaphoreWithPriority` and `AbortController` broke once Vue proxied them, which `markRaw` fixes. Each fix is its own commit.

Lint, vitest and the build pass, `vue-tsc` goes from 347 errors to 108. Playwright, full suite locally: green. Not covered by a spec: upload through the picker and the folders view, the sidebar and dashboard entry points.


skjnldsv:
Will try to split this one a tad further too, brace yourself!! :eyes:

